PR #1, opened 2020-09-08
Rm sign in component

rhystmills:
## What does this change?
This introduces the new 'SignInComponent'. This is a Braze-delivered web message that will appear on TheGuardian.com to digital subscribers who have never signed in to the Guardian app. The component requires some more work when the design is finalised, but the basic layout is ready, as well as button functionality.

Interacting with Braze will happen in both [frontend](https://github.com/guardian/frontend/) and [dotcom-rendering](https://github.com/guardian/dotcom-rendering), with this repository providing the presentation layer.

I've also added support for Guardian fonts in our Storybook setup.

## How to test
Running `yarn storybook` in this repository will show the component in an instance of Storybook in your browser. Running the component in the context of Frontend or DCR is more complicated - follow the instructions [here](https://github.com/guardian/braze-components).

The design of the component can be tested in Storybook, while testing in-situ can be conducted in frontend and DCR using the previously linked readme.

## Images

image


index.ts
```diff
@@ -1 +1,2 @@
-export * from "./src/ExampleComponent";
+// export * from "./src/ExampleComponent";
```
tjmw, line 1: I think let's remove the `ExampleComponent` completely now we've got a real version of the banner. What do you reckon?

index.ts
```diff
@@ -1 +1,2 @@
...
+export * from "./src/SignInComponent";
```
tjmw, line 2: I wonder if there's a different name we could use for this component? `SignInComponent` sounds quite generic as is. Maybe `PremiumSubscriberAppBanner` or something similar?

src/SignInComponent.tsx
```diff
@@ -0,0 +1,342 @@
+import React, { useState } from 'react';
+import { css } from "@emotion/core";
+import { body, headline } from '@guardian/src-foundations/typography/cjs';
+import { palette } from "@guardian/src-foundations";
+import { Button } from "@guardian/src-button";
+import { from, until } from '@guardian/src-foundations/mq';
+import { space } from '@guardian/src-foundations';
+import { SvgCross } from '@guardian/src-icons';
+import { SvgInfo } from '@guardian/src-icons';
```
tjmw, line 9: These could be combined into a single import:

```js
import { SvgCross, SvgInfo } from '@guardian/src-icons';
```

I think it's clearer that way, what do you reckon?

(This is also the case with the `src-foundation` import.)

src/SignInComponent.tsx
```diff
@@ -0,0 +1,342 @@
...
+    box-sizing: inherit;
+  }
+  box-sizing: border-box;
+  display: flex;
+  flex-direction: column;
+  justify-content: center;
+  align-items: center;
+  width: 100%;
+  background-color: ${palette.background.secondary};
+  color: ${"#333"};
```
tjmw, line 37: Is there a palette colour value we could use here?

src/SignInComponent.tsx
```diff
@@ -0,0 +1,342 @@
...
+    padding: ${space[4]}px;
+    ${from.tablet} {
+        width: 55%;
+        padding-right: 0;
+    }
+    ${from.desktop} {
+        width: 60%;
+    }
+    ${from.wide} {
+        width: 60%;
```
tjmw, line 73: Is this needed given that we have the same rule in `from.wide` above?

src/SignInComponent.tsx
```diff
@@ -0,0 +1,342 @@
...
+        max-width: 45%;
+    }
+`;
+
+export const heading = css`
+    ${headline.xsmall({ fontWeight: 'bold' })};
+    margin: 0;
+    max-width: 100%;
+
+    @media (min-width: 740px) {
```
tjmw, line 110: Is there a named breakpoint from source we could use here?

src/SignInComponent.tsx
```diff
@@ -0,0 +1,342 @@
...
+    display: inline-block;
+`
+
+export const smallRightSpacer = css`
+    margin-right: ${space[3]}px;
+`
+
+export const primaryButton = css `
+    margin-right: ${space[3]}px;
+    background-color: #333;
```
tjmw, line 164: Same question about whether there are named palette values we could use here?

src/SignInComponent.tsx
```diff
@@ -0,0 +1,342 @@
...
+
+    const onCloseClick = (evt: React.MouseEvent<HTMLButtonElement, MouseEvent>, buttonId: string): void => {
+        evt.preventDefault();
+        onButtonClick(buttonId);
+        setShowBanner(false);
+    };
+
+    return (
+        <>
+            {showBanner ? (
```
tjmw, line 297: What do you think about replacing this ternary with an early return when `showBanner` is false? I think that would be slightly more clear/readable.

tjmw (commented):
Nice! This looks really great 🏅 I left several comments and suggestions but they're all pretty minor. Happy to pair on or discuss anything that isn't clear.

It looks like `package-lock.json` got committed, which I don't think we should add (likely from running `npm install` locally instead of `yarn install`).

I would possibly think about breaking the main banner component up into sub components a bit to make it easier to read, but certainly not a big issue (and we could attempt a refactor in a separate PR if that's preferable).

src/SignInComponent.tsx
rhystmills: That colour came from the mockup provided by UX, but I couldn't find it in any palettes from Source - would you generally go with the colour provided by UX or find something similar from the palettes?

tjmw: I think my first thought would be to see if there's something similar we can use in the source palette. It's definitely okay to use custom colours (though I'd vote to pull them out and give them a name) but nice if we can use source colours for consistency!

src/SignInComponent.tsx
rhystmills: Yes - I'll switch it over to the tablet breakpoint.

src/SignInComponent.tsx
rhystmills: Good shout, will do!

src/SignInComponent.tsx
rhystmills: I managed to dig one out of Source in the end anyway - included now.

src/SignInComponent.tsx
rhystmills: No - it's gone now.

src/SignInComponent.tsx
rhystmills: Yes good idea, I'll tidy it up a bit.

index.ts
rhystmills: Yes, makes sense to me.

src/DigitalSubscriberAppBanner.tsx
```diff
@@ -0,0 +1,332 @@
...
+                    <div css={packShot}>
+                        <img
+                            src="https://via.placeholder.com/400x300"
+                            alt=""
+                        />
+                    </div>
+                    <div css={iconPanel}>
+                        <button
+                            aria-label="Close"
+                            onClick={e => onCloseClick(e, '1')}
```
paulbrown1982, line 322: What would be the impact of giving this button ID '2'?

rhystmills: I'm not sure actually. '0' maps to 'Button 1' in the Braze interface and '1' maps to 'Button 2'. I'll give it a go and see.

src/DigitalSubscriberAppBanner.tsx
```diff
@@ -0,0 +1,332 @@
...
+    width: 100%;
+    max-height: 100%;
+
+    ${from.tablet} {
+        align-self: flex-end;
+        max-width: 40%;
+        padding-right: ${space[4]}px;
+        padding-right: ${space[4]}px;
+    }
+
```
adamnfish, line 84: This is minor to the point of irrelevance, but should we choose a code style for these blocks? Above, the breakpoints are all smooshed together but we have new lines between each breakpoint here.

I'm not fussed in the slightest, just sharing something I've noticed.

src/DigitalSubscriberAppBanner.tsx
```diff
@@ -0,0 +1,332 @@
...
+    height: 10px;
+    svg {
+        height: 30px;
+        width: auto;
+        margin-top: ${space[2]}px;
+        padding-right: ${space[2]}px;
+    }
+    display: inline-block;
+    vertical-align: -37.5%;
+
```
adamnfish, line 271: Couple of empty lines in funny places like this. Again, I'm not fussed in the slightest though :-)

src/DigitalSubscriberAppBanner.stories.tsx
```diff
@@ -0,0 +1,24 @@
...
+                console.log("Button clicked!");
+            }}
+            message="Test message"
+            firstName="John"
+        />
+    </StorybookWrapper>
+  );
+};
+
+defaultStory.story = { name: "Sign In Component" };
```
adamnfish, line 24: Looks like this name might be copied from a different component?

src/DigitalSubscriberAppBanner.stories.tsx
```diff
@@ -0,0 +1,24 @@
...
+export default {
+  component: DigitalSubscriberAppBanner,
+  title: "Components/DigitalSubscriberAppBanner",
+};
+
+export const defaultStory = (): ReactElement => {
+  return (
+    <StorybookWrapper>
+        <DigitalSubscriberAppBanner
+            onButtonClick={() => {
```
adamnfish, line 14: I'd consider logging the button ID here as well, but not urgent and no harm here. We can add it later if we want to do that.

adamnfish (approved):
Fantastic!

src/DigitalSubscriberAppBanner.tsx
```diff
@@ -0,0 +1,332 @@
...
+    display: flex;
+    justify-content: center;
+    width: 100%;
+    max-height: 100%;
+
+    ${from.tablet} {
+        align-self: flex-end;
+        max-width: 40%;
+        padding-right: ${space[4]}px;
+        padding-right: ${space[4]}px;
```
paulbrown1982, line 82: `padding-right` is repeated

src/DigitalSubscriberAppBanner.tsx
rhystmills: It seems to be coming through as a 'Body Click' when the index is 2. My guess is 0 -> 'Button 1', 1 -> 'Button 2', everything else -> 'Body Click'

wojtek-tg:
As a person with around 0 frontend knowledge - do we need 17k lines of `package-lock.json` ?

src/DigitalSubscriberAppBanner.tsx
rhystmills: Yes, definitely worth going for something consistent here. I've reorganised, going for a couple of line breaks before nested blocks.

src/DigitalSubscriberAppBanner.tsx
rhystmills: They have now been ousted.

src/DigitalSubscriberAppBanner.stories.tsx
rhystmills: Good spot - I renamed the component today and missed that one.

src/DigitalSubscriberAppBanner.tsx
rhystmills: Cheers Paul, fixed now!

src/DigitalSubscriberAppBanner.stories.tsx
rhystmills: Yes, good idea - I've implemented now.

PR #3, opened 2020-09-09
Dynamic fields improvements

rhystmills:
## What does this change?
An update to the dynamic text field which will have more relevance to the end goal (a message into which marketing can input).

Also a small improvement to our JSX indentation.

## How to test
Run `yarn storybook` and check the component in-situ

### DIFF
--- a/src/DigitalSubscriberAppBanner.tsx
+++ b/src/DigitalSubscriberAppBanner.tsx
@@ -12,8 +12,8 @@ const imgHeight = "300";
 
 export type Props = {
   onButtonClick: (buttonType: string) => void;
-  header?: string;
-  body?: string;
+  header: string;
+  body: string;
 };
 
 export const wrapper = css`
--- a/src/DigitalSubscriberAppBanner.tsx
+++ b/src/DigitalSubscriberAppBanner.tsx
@@ -306,9 +306,12 @@ export const DigitalSubscriberAppBanner: React.FC<Props> = ({
                     <div css={infoIcon}>
                         <SvgInfo />
                     </div>
-                    <div css={heading}><span css={smallInfoIcon}>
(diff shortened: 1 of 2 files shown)


src/DigitalSubscriberAppBanner.tsx
```diff
@@ -11,9 +11,9 @@ import { PlayStore } from "./assets/play-store"
 const imgHeight = "300";
 
 export type Props = {
-  message: string;
   onButtonClick: (buttonType: string) => void;
-  firstName?: string;
+  header?: string;
+  body?: string;
```
tjmw, line 16: The `?` makes these props optional, but I think they should be required.

tjmw (approved):
Nice, this looks good to me!

Chromatic is flagging up a visual difference (because we went from "Hi" to "Hello" in the body text. I think it's fine to click through to Chromatic and accept that change.

src/DigitalSubscriberAppBanner.tsx
rhystmills: I'd left them optional as they could be left blank in the Braze interface, which would perhaps yield a null? What do you think?

tjmw: Hmm good question. They both seem like required fields to me in that I don't think the banner makes sense without either of them.

Though it's true that these TypeScript types don't give us any kind of enforcement that they'll actually be set at runtime. If that's a concern then perhaps the right approach is to leave them optional but return early if either aren't set. I don't think we'd want to continue with the rendering of the banner if the body text or heading was missing. What do you reckon?

rhystmills: Good idea, I've kept them optional but added a check that will check if they're empty.

PR #4, opened 2020-09-09
Add Guardian breakpoints to Storybook

tjmw:
## What does this change?

Adds easy switching between source viewport widths to Storybook (viewport config taken from the support-dotcom-components project).


rhystmills (approved):
Looks great! Nothing to add.

PR #5, opened 2020-09-10
Buttons IDs and property name consistency

rhystmills:
## What does this change?
I've changed the integers in strings used as button IDs to be pure numbers - once called in Frontend or DCR, the InAppMessageButtonClick constructor expects a number, so providing a number-typed value here saves us work in the repositories where the component is used.

I've also updated the case of the property names displayed in Storybook (i.e. "header", "body") to be lower case, in order to exactly match their names in the Braze interface.

## How to test
The component can be tested in Storybook using `yarn storybook`.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@guardian/braze-components",
-  "version": "0.0.5",
+  "version": "0.0.6",
   "description": "React components to render messages from Braze",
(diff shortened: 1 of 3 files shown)


paulbrown1982 (approved):
👍  LGTM

tjmw (approved):
Nice work!

PR #6, opened 2020-09-14
Accessibility improvements

rhystmills:
## What does this change?
The `closeButton` is now keyboard focusable, showing the focusHalo when navigated to using the tab key. For now, the tabindex has been set to 0, but could be increased based on feedback from the E&P/Accessibility group.

There is an open question of how the banner should be navigated to - screen-reader users may find the banner unobtrusive as it is (because it's sequentially after the main page content), while users navigating by keyboard may want to close it without tabbing through the rest of the page.

## How to test
View in storybook by running `yarn storybook`.

image

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@guardian/braze-components",
-  "version": "0.0.6",
+  "version": "0.0.7",
   "description": "React components to render messages from Braze",
   "repository": "https://github.com/guardian/braze-components",
   "author": "The Guardian",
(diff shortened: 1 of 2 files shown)


tjmw (approved):
Nice work! I think this looks good. Really interesting point with regard to the open question you mention in the description. I don't think we should let that stop us merging this PR. Would be good to see what feedback you get in the accessibility channel on the question.

As an aside - I noticed the version bump happened in this branch. What do you think of always bumping and releasing from the `main` branch? There's a bit of a possible race condition with  bumping on a branch where two developers branch off of main at `0.0.6` and then bump on a branch (though I'm pretty sure NPM would stop you releasing two different `0.0.7` releases).

PR #7, opened 2020-09-17
Changes related to new design

rhystmills:
## What does this change?
- Palette and font changes in response to new designs.
- Modifications to Tablet layout - now similar to Phablet layout (i.e. no image) as the image was too cramped in this layout.
- Addition of screenshot from Grid.
- Close button now based on source component, rather than custom button build from source svg (focusHalo no longer needs to be explicitly added for focus)
- Addition of GT Titlepiece font to the repo in order to support future use.

## How to test
Run `yarn storybook` to see the component in situ.

## Images
****
image
image
image

### DIFF
--- a/src/DigitalSubscriberAppBanner.tsx
+++ b/src/DigitalSubscriberAppBanner.tsx
@@ -1,15 +1,18 @@
 import React, { useState } from 'react';
 import { css } from "@emotion/core";
+import { ThemeProvider } from 'emotion-theming'
 import { palette, space } from "@guardian/src-foundations";
 import { from, until } from '@guardian/src-foundations/mq';
 import { body, headline } from '@guardian/src-foundations/typography/cjs';
-import { Button } from "@guardian/src-button";
+import { Button, buttonReaderRevenueBrandAlt } from "@guardian/src-button";
 import { SvgCross, SvgInfo } from '@guardian/src-icons';
 import { AppStore } from "./assets/app-store";
 import { PlayStore } from "./assets/play-store";
-import { focusHalo } from "@guardian/src-foundations/accessibility";
 
-const imgHeight = "300";
+const imgHeight = "280";
+const bannerColor = "#ebe8e8";
+const infoColor = "#c4c4c4";
+const bodyColor = "#666";
 
 export type Props = {
   onButtonClick: (buttonIndex: number) => void;
@@ -24,7 +27,7 @@ export const wrapper = css`
   justify-content: center;
   align-items: center;
   width: 100%;
-  background-color: ${palette.background.secondary};
+  background-color: ${bannerColor};
   color: ${palette.neutral[20]};
 
   html {
@@ -45,13 +48,11 @@ export const contentContainer = css`
     position: relative;
     margin: 0 auto;
     width: 100%;
-    max-width: 980px;
-
-    ${from.tablet} {
-        flex-direction: row;
-    }
+    max-width: 740px;
 
     ${from.desktop} {
+        max-width: 980px;
+        flex-direction: row;
         min-height: ${imgHeight}px;
     }
 
@@ -68,11 +69,6 @@ export const topLeftComponent = css`
     width: 93%;
     padding: ${space[4]}px;
 
-    ${from.tablet} {
-        width: 55%;
-        padding-right: 0;
-    }
-
     ${from.desktop} {
         width: 60%;
     }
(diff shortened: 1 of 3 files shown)


src/DigitalSubscriberAppBanner.tsx
```diff
@@ -248,10 +222,11 @@ export const smallInfoIcon = css`
         width: 40px;
         height: 40px;
         vertical-align: -30%;
-        fill: ${palette.neutral[86]};
+        fill: #c4c4c4;
```
adamnfish, line 225: Should this be `infoColor`?

tjmw (approved):
🏅 Great work!

src/DigitalSubscriberAppBanner.tsx
```diff
@@ -266,11 +241,17 @@ export const infoIcon = css`
...
+    
     svg {
         width: 60px;
         height: 60px;
-        fill: ${palette.neutral[86]};
+        fill: ${infoColor};
+        background: radial-gradient(circle at center, ${palette.background.primary} 0%, ${palette.background.primary} 50%, ${palette.background.primary} 52%, transparent 53%);
+        border-radius: 50%;
+        
+        path {
```
adamnfish, line 252: Does this need to be here?

src/DigitalSubscriberAppBanner.tsx
```diff
@@ -335,19 +316,23 @@ export const DigitalSubscriberAppBanner: React.FC<Props> = ({
                 <div css={bottomRightComponent}>
                     <div css={packShot}>
                         <img
-                            src="https://via.placeholder.com/400x300"
+                            src="https://media.guim.co.uk/86d117fc42742412c11918cdce82f47c83f65881/32_113_812_487/812.png"
```
adamnfish, line 319: Are we happy to hotlink this, rather than including it as part of the component and bundle, directly? I think this is fine, but just checking.

adamnfish (approved):
LGTM. Couple of comments but nothing important.

yarn.lock
```diff
@@ -5627,7 +5627,7 @@ emojis-list@^3.0.0:
   resolved "https://registry.yarnpkg.com/emojis-list/-/emojis-list-3.0.0.tgz#5570662046ad29e2e916e71aae260abdff4f6a78"
   integrity sha512-/kyM18EfinwXZbno9FyUGeFh87KC8HRQBQGildHZbEuRyWFOmv1U10o9BBp8XVZDVNNuQKyIGIu5ZYAAXJ0V2Q==
 
-emotion-theming@^10.0.19:
+emotion-theming@^10.0.19, emotion-theming@^10.0.27:
```
paulbrown1982, line 5630: Do we still need `emotion-theming@^10.0.19` in here?

src/DigitalSubscriberAppBanner.tsx
rhystmills: Definitely not, have removed!

src/DigitalSubscriberAppBanner.tsx
rhystmills: Support-frontend have a system for getting images from Grid [here](https://github.com/guardian/support-frontend/blob/920e638c35430cc260acdb1878f37bffa1d12fae/support-frontend/assets/helpers/theGrid.js) - as far as I can tell it ultimately assembles one of these media.guim urls. I did try adding the image into the component as a relative link but it didn't seem to be working in Storybook.

tjmw: One other thing to add. The reader revenue banners specify the image slightly differently:

`https://i.guim.co.uk/img/media/773ead1bd414781052c0983858e6859993870dd3/34_72_1825_1084/1825.png?width=500&quality=85&s=24cb49b459c52c9d25868ca20979defb`

I'm not sure of the difference between i.guim and media.guim, but perhaps the fact that rr are using i.guim is something to learn from and we should do the same? They're also controlling the size/quality via query string args. I _think_ the `s` is a signature for the URL so it can't be tampered with, but I'm not certain how we'd generate that. Asking in the Dotcom channel would probably get some good answers!

yarn.lock
rhystmills: I think 10.0.19 is required by some of the other packages, I had added 10.0.27 for direct use by us. I think it would be easier to stick with 10.0.19

src/DigitalSubscriberAppBanner.tsx
rhystmills: I've asked on Dotcom, it would be good to know!

paulbrown1982: i.guim contains Fastly's image resizing. So if you don't know your exact crop size you can let that system do it. i.guim is probably more useful for JPEG photos rather than PNG graphics.

rhystmills: Excellent. I've got it on i.guim now, shaving off a little bit of file size

PR #8, opened 2020-09-22
Get eslint & prettier working consistently

tjmw:
## What does this change?

I noticed that the formatting my editor was applying was different than the recent commits made by Rhys. It turned out that we didn’t have eslint and prettier config for this repo. Added and reformatted the main component and story. Slightly annoying to have to do this because the diff is noisy, but better now than down the line when we're supporting more components.

There should be no user-visible changes from this PR.

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,22 @@
+; This file is for unifying the coding style for different editors and IDEs
+; See editorconfig.org
+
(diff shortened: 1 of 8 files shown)


adamnfish (approved):
All looks fine to me, thank for doing this in a separate PR 👍

PR #9, opened 2020-09-22
Specify component name as a prop

tjmw:
## What does this change?

Adds support for specifying the specific component to render as a prop which is passed to the (new) generic `BrazeMessage` component:

```js
<BrazeMessage componentName="DigitalSubscriberAppBanner" ... />
```

The component to render will be configured in Braze by marketing. The `BrazeMessage` includes a mappings/allow list of valid component names.

The platforms will be updated to import `BrazeMessage`, threading through other details (including the component name) from Braze.

Note that this PR adds Jest support for the first time, and there are some supporting changes around that. I've also setup a GitHub action to run the jest tests as part of this PR.


src/BrazeMessage.tsx
```diff
@@ -0,0 +1,48 @@
...
+
+const COMPONENT_MAPPINGS: ComponentMapping = {
+    DigitalSubscriberAppBanner,
+};
+
+export type Props = {
+    onButtonClick: (buttonIndex: number) => void;
+    componentName: string;
+    brazeMessageProps: {
+        [key: string]: string;
```
adamnfish, line 25: What's the reason for not using the `BrazeMessageProps` type here?

src/BrazeMessage.test.tsx
```diff
@@ -0,0 +1,47 @@
...
+import { buildBrazeMessageComponent } from './BrazeMessage';
+
+describe('BrazeMessage', () => {
+    it('renders the correct component when a valid componentName is passed', () => {
+        const ExampleComponent = jest.fn(() => null);
+        const mappings = {
+            ExampleComponent,
+        };
+        const BrazeMessage = buildBrazeMessageComponent(mappings);
+        const componentName = 'ExampleComponent';
```
adamnfish, line 16: I'd consider inlining this const in both tests. It's only used once, and it is the key part of what is being tested, so a shame if it vanishes into the noise above? Super minor point and I don't think it matters either way - up to you!

adamnfish (approved):
Looks like a great change to me, thanks! We worked together on some of this, so maybe let's see what fresh eyes say as well.

src/BrazeMessage.tsx
tjmw: Good spot, I just missed this, no reason at all not to use it.

src/BrazeMessage.test.tsx
tjmw: Good point! The reason I'd pulled it out was to give it a name because like you said it's really central to the test. But given that when it's passed as a prop, it also gets that name, I think your suggestion to inline is a good one. I'll make the change :+1:

src/DigitalSubscriberAppBanner.stories.tsx
```diff
@@ -1,26 +1,29 @@
...
     title: 'Components/DigitalSubscriberAppBanner',
     decorators: [withKnobs],
 };
 
 export const defaultStory = (): ReactElement => {
     return (
         <StorybookWrapper>
-            <DigitalSubscriberAppBanner
+            <BrazeMessage
+                componentName={text('componentName', 'DigitalSubscriberAppBanner')}
```
adamnfish, line 16: It'd be magic if this was a drop-down. No harm for now though, this will only matter when there are multiple options.

tjmw: Yep good point. I think okay to leave for a future version though like you suggested. This did remind me to add `slotName` as an (unused) prop control in Storybook, so the fields line up with what's in Braze.

adamnfish: OK yeah, I was gonna chase that up with you separate, but great to do it at the same time for doing a single release, I guess.

PR #10, opened 2020-09-22
Run eslint as a GitHub action

tjmw:
## What does this change?

Runs eslint as a GitHub action.


adamnfish (commented):
Would it be better to add the lint step directly to the test step? This would mean there's no need for a separate action here, it would be caught by the test. It also means we're more likely to catch it locally. Running tests is always part of the dev workflow, so that would eliminate the need to remember to run `yarn lint` as well.

adamnfish (approved):
No harm though, great as-is and definitely excellent to have this check in place on the CI in any form 👍

PR #11, opened 2020-09-23
Make the two main buttons behave the same

tjmw:
## What does this change?

In terms of their impact on moving through the canvas in Braze, we want to treat "OK, got it" and "I'm not interested" the same.


rhystmills (approved):
Breathtaking

PR #12, opened 2020-09-23
Publish Package as an ES Module

tjmw:
## What does this change?

Publish our package as an ES module

This is the [approach we’re settling on in Client Side infra discussions](https://github.com/orgs/guardian/teams/client-side-infra/discussions/11).

Furthermore, without this change the module no longer loads correctly on frontend since guardian/frontend#23011 was merged, which changes the config to transpile all @guardian NPM packages.

## Images

Here's the banner component loading successfully on DCR and frontend (I've forced the component to load, rather than triggering from Braze, which is why the copy is wrong):

Screenshot 2020-09-23 at 16 09 54

Screenshot 2020-09-23 at 16 18 36

### DIFF
--- a/package.json
+++ b/package.json
@@ -5,7 +5,7 @@
   "repository": "https://github.com/guardian/braze-components",
   "author": "The Guardian",
   "license": "MIT",
-  "main": "dist/index.js",
+  "module": "dist/index.js",
   "types": "dist/index.d.ts",
   "files": [
     "dist"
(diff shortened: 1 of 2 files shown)


sndrs (approved):
> the module no longer loads correctly on frontend

😞 sorry!

tjmw:
Thanks for the review @sndrs! No need to apologise, it works now that we're following the client side infra recommendations (also this feature isn't in prod just yet so no user facing impact).

PR #13, opened 2020-09-23
Added css fix for IE11 and Edge

rhystmills:
## What does this change?
A small fix for IE11 and Edge, keeping the primary image in correct dimensions and keeping the app store icons inline.

### DIFF
--- a/src/DigitalSubscriberAppBanner.tsx
+++ b/src/DigitalSubscriberAppBanner.tsx
@@ -70,8 +70,8 @@ export const contentContainer = css`
 export const topLeftComponent = css`
     width: 93%;
     padding: ${space[4]}px;
-    min-height: ${imgHeight};
-    {from.desktop} {
+    min-height: ${imgHeight}px;
+    ${from.desktop} {
         width: 60%;
     }
 `;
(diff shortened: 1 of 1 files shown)


src/DigitalSubscriberAppBanner.tsx
```diff
@@ -70,8 +70,8 @@ export const contentContainer = css`
 export const topLeftComponent = css`
     width: 93%;
     padding: ${space[4]}px;
-
-    ${from.desktop} {
+    min-height: ${imgHeight};
+    {from.desktop} {
```
tjmw, line 74: Should this be `${from.desktop}`?

src/DigitalSubscriberAppBanner.tsx
```diff
@@ -70,8 +70,8 @@ export const contentContainer = css`
 export const topLeftComponent = css`
     width: 93%;
     padding: ${space[4]}px;
-
-    ${from.desktop} {
+    min-height: ${imgHeight};
```
tjmw, line 73: Should we specify a unit here (`px`) for consistency? (I think _technically_ it's not valid css to omit the unit, might be wrong though).

tjmw (approved):
I think this is looking good! I left a couple of comments which it'd be nice to resolve before merging. Maybe add some screenshots from the relevant browsers to the PR description?

src/DigitalSubscriberAppBanner.tsx
rhystmills: Cheers Tom, both fixed now!

PR #14, opened 2020-09-24
Publish package as esm and cjs

tjmw:
## What does this change?

This switches to publishing the package as esm and cjs. I followed the example of the cmp library. When only publishing as esm I ran into an issue where the linter was failing the find the package on the platforms. This seems to fix the issue.

Note that in order to test this I published the package off of this branch (`0.0.10`).


rhystmills (approved):
Looks good to me.

PR #15, opened 2020-09-25
Add Rollup Visualizer Plugin

tjmw:
## What does this change?

This outputs a `stats.html` file when building the package so we can see what our bundle is made up of.

Also includes a tidy up of the rollup config (separate commit).

Screenshot 2020-09-25 at 10 09 20


rhystmills (approved):
Looks great!

PR #16, opened 2020-09-25
Updated README to include more information on testing and updating the NPM module.

rhystmills:
Added some useful info for testing in Frontend regarding `make watch` tracking the linked module.

Also added an instruction for updating `main` when a new version of the NPM module is published.

### DIFF
--- a/README.md
+++ b/README.md
@@ -39,9 +39,12 @@ $ yarn install --force
 
 [`yarn link`]: https://classic.yarnpkg.com/en/docs/cli/link/
 
-NOTE: Ensure you build this library before adding it locally to your project,
+NOTE:
+- Ensure you build this library before adding it locally to your project,
 by running `yarn build`. You can also use `yarn watch` to build automatically
 when the source code is changed.
+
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -58,3 +61,6 @@ This will:
 - Interactively ask for a new version number (and create a commit for the
   version change in package.json)
 - Publish the new release to NPM
+- Push a single commit to your local version of this repository
```
tjmw, line 64: How about updating this to something slightly more explicit?

```suggestion
- Push the newly created version bump commit to GitHub
```

Alternatively, I wonder if instead of explaining this as another manual step, maybe we should push automatically as part of running `yarn publishToNPM`? Maybe let's update the instruction for now and mull over automating it (maybe there's a reason not to instantly push 🤔)

tjmw (approved):
Nice 🎖️ I think this looks good. I left one minor suggestion to take or leave.

README.md
rhystmills: That does make sense, I'll update the wording for now to be a bit clearer, and we can look into automatic pushing soon.

PR #17, opened 2020-09-30
Handle click tracking in the component

tjmw:
## What does this change?

The platform is expected to provide an implementation of `submitComponentEvent` to the component as a prop. We can then use this whenever we want to add Ophan tracking in the component. Initially we'll be using it for tracking clicks. This This reduces duplication and arguably makes it easier to test.

Note that I'm pointing to a GitHub branch for `@guardian/types` to get the Ophan types. When (if!) that branch is merged we can switch to getting from NPM.

### DIFF
--- a/package.json
+++ b/package.json
@@ -29,6 +29,7 @@
     "@babel/preset-react": "^7.10.4",
     "@babel/preset-typescript": "^7.10.4",
     "@emotion/babel-preset-css-prop": "^10.0.27",
+    "@guardian/types": "^0.4.5",
     "@rollup/plugin-alias": "^3.1.1",
     "@rollup/plugin-babel": "^5.1.0",
     "@rollup/plugin-commonjs": "^14.0.0",
--- a/src/BrazeMessage.tsx
+++ b/src/BrazeMessage.tsx
@@ -1,12 +1,18 @@
 import React from 'react';
-import { DigitalSubscriberAppBanner } from './DigitalSubscriberAppBanner';
+import { OphanComponentEvent } from '@guardian/types/ophan';
+import {
+    COMPONENT_NAME as DIGITAL_SUBSCRIBER_APP_BANNER_NAME,
+    DigitalSubscriberAppBanner,
+} from './DigitalSubscriberAppBanner';
+import { BrazeClickHandler } from './tracking';
 
 type BrazeMessageProps = {
     [key: string]: string | undefined;
 };
 
 type CommonComponentProps = {
-    onButtonClick: (buttonIndex: number) => void;
+    logButtonClickWithBraze: BrazeClickHandler;
+    submitComponentEvent: (componentEvent: OphanComponentEvent) => void;
     brazeMessageProps: BrazeMessageProps;
 };
 
@@ -15,17 +21,23 @@ type ComponentMapping = {
 };
 
 const COMPONENT_MAPPINGS: ComponentMapping = {
-    DigitalSubscriberAppBanner,
+    [DIGITAL_SUBSCRIBER_APP_BANNER_NAME]: DigitalSubscriberAppBanner,
 };
 
 export type Props = {
-    onButtonClick: (buttonIndex: number) => void;
+    logButtonClickWithBraze: BrazeClickHandler;
+    submitComponentEvent: (componentEvent: OphanComponentEvent) => void;
     componentName: string;
     brazeMessageProps: BrazeMessageProps;
 };
 
 export const buildBrazeMessageComponent = (mappings: ComponentMapping): React.FC<Props> => {
-    const BrazeMessageComponent = ({ onButtonClick, componentName, brazeMessageProps }: Props) => {
+    const BrazeMessageComponent = ({
+        logButtonClickWithBraze,
+        submitComponentEvent,
+        componentName,
+        brazeMessageProps,
+    }: Props) => {
         const ComponentToRender = mappings[componentName];
 
         if (!ComponentToRender) {
@@ -34,7 +46,8 @@ export const buildBrazeMessageComponent = (mappings: ComponentMapping): React.FC
 
         return (
             <ComponentToRender
-                onButtonClick={onButtonClick}
+                logButtonClickWithBraze={logButtonClickWithBraze}
+                submitComponentEvent={submitComponentEvent}
                 brazeMessageProps={brazeMessageProps}
             />
         );
--- a/src/DigitalSubscriberAppBanner.test.tsx
+++ b/src/DigitalSubscriberAppBanner.test.tsx
@@ -1,0 +1,83 @@
+/**
+ * @jest-environment jsdom
+ */
+
+import React from 'react';
+import { render, fireEvent } from '@testing-library/react';
+import { DigitalSubscriberAppBanner } from './DigitalSubscriberAppBanner';
+
+describe('DigitalSubscriberAppBanner', () => {
+    describe('when a button is clicked', () => {
+        const baseProps = () => ({
+            logButtonClickWithBraze: jest.fn(),
+            submitComponentEvent: jest.fn(),
+            brazeMessageProps: {
+                header: 'A note to our digital subscribers',
+                body: 'Hi John, did you know...',
+            },
+        });
+
+        it('invokes logButtonClickWithBraze with the internal button ID', () => {
+            const logButtonClickWithBraze = jest.fn();
+            const { getByText } = render(
+                <DigitalSubscriberAppBanner
+                    {...baseProps()}
+                    logButtonClickWithBraze={logButtonClickWithBraze}
+                />,
+            );
+
+            fireEvent.click(getByText('Ok, got it'));
+
+            expect(logButtonClickWithBraze).toHaveBeenCalledWith(0);
+        });
+
+        it('invokes submitComponentEvent with correct data', () => {
+            const logButtonClickWithOphan = jest.fn();
+            const { getByText } = render(
+                <DigitalSubscriberAppBanner
+                    {...baseProps()}
+                    submitComponentEvent={logButtonClickWithOphan}
+                />,
+            );
+
+            fireEvent.click(getByText('Ok, got it'));
+
+            expect(logButtonClickWithOphan).toHaveBeenCalledWith({
+                component: {
+                    componentType: 'RETENTION_ENGAGEMENT_BANNER',
+                    id: 'DigitalSubscriberAppBanner',
+                },
+                action: 'CLICK',
+                value: '1',
+            });
+        });
+
+        it('closes the banner', () => {
+            const props = baseProps();
+            const { container, getByText } = render(<DigitalSubscriberAppBanner {...props} />);
+
+            fireEvent.click(getByText('Ok, got it'));
+
+            // Closing the banner means nothing is rendered
+            expect(container.firstChild).toBeNull();
+        });
+
+        it('closes the banner even if the callback throws an error', () => {
+            const props = baseProps();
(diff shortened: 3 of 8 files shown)


package.json
```diff
@@ -29,6 +29,7 @@
     "@babel/preset-react": "^7.10.4",
     "@babel/preset-typescript": "^7.10.4",
     "@emotion/babel-preset-css-prop": "^10.0.27",
+    "@guardian/types": "guardian/types#tw-add-ophan-types",
```
tjmw, line 32: When this branch is merged we can create a new release of `@guardian/types` and use that.

src/DigitalSubscriberAppBanner.tsx
```diff
@@ -284,18 +287,34 @@ export const storeIcon = css`
...
+
         setShowBanner(false);
+
+        // Braze displays button id from 1, but internal representation is numbered from 0
+        // This means that the Button ID in Braze and Ophan will be the same
+        const externalButtonId = internalButtonId + 1;
+        submitComponentEvent({
+            component: {
+                // TODO: Fix this, the types need updating:
+                componentType: 'ACQUISITIONS_ENGAGEMENT_BANNER',
```
tjmw, line 310: Note that this needs updating. We want to be using `RETENTION_ENGAGEMENT_BANNER`, which is yet to be added as a new component type.

src/DigitalSubscriberAppBanner.test.tsx
```diff
@@ -0,0 +1,83 @@
...
+            const props = baseProps();
+            const { getByText, queryByText } = render(<DigitalSubscriberAppBanner {...props} />);
+
+            fireEvent.click(getByText('Ok, got it'));
+
+            expect(queryByText(props.brazeMessageProps.header)).toBe(null);
+            expect(queryByText(props.brazeMessageProps.body)).toBe(null);
+        });
+
+        it('handles errors in the callbacks', () => {
```
adamnfish, line 65: Not crystal clear to me here what we mean by "handles"? By comparing with the above it looks like the banner gets closed when an error happens? If that's the behaviour we want is it worth mentioning that in the test string?

src/DigitalSubscriberAppBanner.test.tsx
```diff
@@ -0,0 +1,83 @@
...
+            });
+        });
+
+        it('closes the banner', () => {
+            const props = baseProps();
+            const { getByText, queryByText } = render(<DigitalSubscriberAppBanner {...props} />);
+
+            fireEvent.click(getByText('Ok, got it'));
+
+            expect(queryByText(props.brazeMessageProps.header)).toBe(null);
```
adamnfish, line 61: It's not obvious to me as someone inexperienced with React that checking for null properties is the same as checking the component has been hidden/removed. If this is obvious to Reactors then no worries at all, but if not is there a more direct way to check if the component is no long visible?

src/DigitalSubscriberAppBanner.tsx
```diff
@@ -283,19 +286,45 @@ export const storeIcon = css`
     }
 `;
 
+const withTryCatchBehaviour = (fn: () => void): void => {
+    try {
+        fn();
+    } catch (e) {}
```
adamnfish, line 292: It's always a huge red flag to me to see errors being silently swallowed like this! Should we not be doing something with these errors? Printing to the console in a pinch, and ideally sending to Sentry. I feel like this is something we will regret in the future when it becomes difficult to debug inconsistent problems and strange behaviour.

I'd also observe that this says `withTryCatchBehaviour`, but by not providing the ability to recover this isn't really giving us try/catch behaviour at all. This might be more accurately named as something like "swallowErrors"?

src/DigitalSubscriberAppBanner.tsx
```diff
@@ -283,19 +286,45 @@ export const storeIcon = css`
...
         evt: React.MouseEvent<HTMLButtonElement, MouseEvent>,
-        buttonId: number,
+        internalButtonId: number,
     ): void => {
         evt.preventDefault();
-        onButtonClick(buttonId);
+
         setShowBanner(false);
+
+        withTryCatchBehaviour(() => {
```
adamnfish, line 310: I'm not sure that the indirection of the `withTryCatchBehaviour` function is really giving us anything here. Is there a reason not to use try / catch directly here and below? This might also give us the opportunity to make better local decisions about how to recover from errors instead of silently discarding them as we do now.

src/DigitalSubscriberAppBanner.tsx
```diff
@@ -283,19 +286,45 @@ export const storeIcon = css`
...
+            // Braze displays button id from 1, but internal representation is numbered from 0
+            // This means that the Button ID in Braze and Ophan will be the same
+            const externalButtonId = internalButtonId + 1;
+            submitComponentEvent({
+                component: {
+                    // TODO: Fix this, the types need updating:
+                    componentType: 'RETENTION_ENGAGEMENT_BANNER' as OphanComponentType,
+                    id: 'DigitalSubscriberAppBanner',
```
adamnfish, line 318: We've a point of fragility here, ensuring that this string stays in sync with the mappings. Can we put this hard coded string in one place to make ourselves a bit more resilient to changes in future?

src/BrazeMessage.tsx
```diff
@@ -19,13 +22,19 @@ const COMPONENT_MAPPINGS: ComponentMapping = {
...
     componentName: string;
     brazeMessageProps: BrazeMessageProps;
 };
 
 export const buildBrazeMessageComponent = (mappings: ComponentMapping): React.FC<Props> => {
-    const BrazeMessageComponent = ({ onButtonClick, componentName, brazeMessageProps }: Props) => {
+    const BrazeMessageComponent = ({
+        logButtonClickWithBraze,
+        submitComponentEvent,
+        componentName,
```
adamnfish, line 35: I wonder if this `CommonComponentProps` vs `Props` distinction could be simplified away, by treating the `componentName` as a separate argument rather than a property. This is totally off topic for the PR though so let's chat separately. It has come up because these types made me go a bit cross-eyed while trying to work out what's happening here.

src/DigitalSubscriberAppBanner.test.tsx
```diff
@@ -0,0 +1,83 @@
+/**
+ * @jest-environment jsdom
+ */
+
+import React from 'react';
+import { render, fireEvent } from '@testing-library/react';
+import { DigitalSubscriberAppBanner } from './DigitalSubscriberAppBanner';
+
+describe('DigitalSubscriberAppBanner', () => {
```
adamnfish, line 9: These are great!

src/DigitalSubscriberAppBanner.test.tsx
```diff
@@ -0,0 +1,83 @@
...
+    describe('when a button is clicked', () => {
```
adamnfish, line 10: I guess we'll test the impression event separately?

src/DigitalSubscriberAppBanner.tsx
```diff
@@ -283,19 +286,45 @@ export const storeIcon = css`
...
+        internalButtonId: number,
     ): void => {
         evt.preventDefault();
-        onButtonClick(buttonId);
+
         setShowBanner(false);
+
+        withTryCatchBehaviour(() => {
+            // Braze displays button id from 1, but internal representation is numbered from 0
+            // This means that the Button ID in Braze and Ophan will be the same
```
adamnfish, line 312: `s/means/ensures/`, perhaps?

src/DigitalSubscriberAppBanner.test.tsx
tjmw: Yeah you're right, being really explicit in the description would be helpful here. Basically the thing I was interested in making more robust here is that even if the tracking throws an error the banner still gets closed.

src/DigitalSubscriberAppBanner.test.tsx
tjmw: This is a good point. I know from experience having tried to do this in the past that it isn't trivial to expect that nothing was rendered, which is basically what we're after here. Testing for the lack of header and body was an imperfect proxy for that. I'll dig a bit more, maybe I've missed something.

src/DigitalSubscriberAppBanner.tsx
tjmw: Good shout. I agree that ideally we want to be passing through an error handler or similar from the platforms (I commented on this in the commit message, but that wasn't very visible - apologies!) I think rather than do too much in a single PR console logging the error would be a good compromise :+1:

src/DigitalSubscriberAppBanner.tsx
tjmw: I think if we switch to logging the errors, which for now would be the common way of handling them, then I can just about justify the existence of this function. I don't feel strongly though, it was mostly just that `try/catch` repeated looked a little untidy to me as it adds so many extra lines!

src/DigitalSubscriberAppBanner.tsx
tjmw: Ah great spot. Yep I'd meant to come back to this.

src/DigitalSubscriberAppBanner.test.tsx
tjmw: This is an interesting comment! As of this PR the _impression_ is still handled entirely by the platform. I've been mulling over the best way of handling this. I think if we want to move to the component, as we're doing with clicks, then a separate PR would be good.

One thing that's got me not 100% convinced about this is that in the case of a banner rendering and VIEWing are 1:1, but for other slots that might not be the case. For example an epic VIEW only happens when it enters the viewport potentially after it's been rendered. It may never be viewed at all. I don't know if the component would have enough information to make that decision, I'd like to try it out. But I'd rather not set a pattern without testing it if that pattern isn't going to work for us in future scenarios.

src/DigitalSubscriberAppBanner.tsx
adamnfish: I wonder if a `finally` block would be a better fit to ensure that we can perform some logic whether or not the handler fails?

Even re-throwing the error should be fine once we've done our cleanup work, so that sentry handles it as normal.

src/DigitalSubscriberAppBanner.test.tsx
adamnfish: No harm if this is the best way, but going on what you've said here it might be worth adding a comment that summarises what you have said above?

src/DigitalSubscriberAppBanner.test.tsx
adamnfish: I like the reasoning here 👍

src/DigitalSubscriberAppBanner.test.tsx
tjmw: Addressed in 4a64e5ae39408f529c86a14bc73bcb0c59571ca6

src/DigitalSubscriberAppBanner.tsx
tjmw: I've started logging any caught errors (and changed the name of the wrapper function, which I've kept, to be clearer): d73ae90197a3da0f45707ba4be279abf2b895707

src/DigitalSubscriberAppBanner.tsx
tjmw: I've split apart the component name from the implementation of the component in 7bf30d52f4180c6b45dc56142b59714c9e6c1fe9 which allows us to use the component name here.

adamnfish (approved):
Looks great once the `guardian/types` update lands, nice one!

PR #25, opened 2020-10-09
Added GuPreview, a mechanism to preview user-populated components on the Guardian site

rhystmills:
## What does this change?
This addon allows users to preview a component in the context of the Guardian website, via a new icon in the header toolbar.

User input in the 'knobs' panel (e.g, 'body' and 'header' copy below) is used to populate the component on the Guardian preview page.

image
### ... That header link goes to :arrow_heading_down:

image

## Additional changes
Storybook has been bumped to version ^6.0.12 in order to support some of its related packages (e.g. storybook/components)

## How to test
Run `yarn storybook` in your local version of this repository and follow the link in the toolbar.


.storybook/gu-preview/GuPreview.tsx
```diff
@@ -0,0 +1,40 @@
+/* eslint-disable no-fallthrough */
```
tjmw, line 1: I'm interested in why we need this. If we do need it is it possibly to scope it more tightly? (using the `eslint-disable-next-line` version perhaps?)

.storybook/gu-preview/GuPreview.tsx
```diff
@@ -0,0 +1,40 @@
...
+
+    const theKnobs = knobsData.get();
+
+    let fullUrl = baseUrl + 'force-braze-message=' + encodeURIComponent(JSON.stringify(theKnobs));
+    window.open(fullUrl);
+}
+
+export const GuPreview = (): ReactElement => {
+    return (
+        <Fragment>
```
tjmw, line 30: I think in general the preferred style at the Guardian is the use the `<>`/`</>` version of fragments (perhaps we should have a rule for this in the eslint config).

.storybook/gu-preview/GuPreview.tsx
```diff
@@ -0,0 +1,40 @@
...
+}));
+
+const IconButtonLabel = styled.div<{}>(({ theme }) => ({
+    fontSize: theme.typography.size.s2 - 1,
+    marginLeft: 10,
+}));
+
+function constructPreviewUrl() {
+    let baseUrl =
+        'https://www.theguardian.com/sport/2019/jul/28/tour-de-france-key-moments-egan-bernal-yellow-jersey?dcr=false&';
```
tjmw, line 20: We actually shouldn't need the `dcr=false` here any more, since forcing is now available on both platforms, no harm though.

tjmw (approved):
💯  This looks great, really nice work figuring it all out (and for the storybook integration looking so seamless!) I left a few minor comments to take or leave.

.storybook/gu-preview/GuPreview.tsx
rhystmills: I'll look at how removing it effects things - it came from the Viewport addon from which I re-used some boiletplate.

.storybook/gu-preview/GuPreview.tsx
rhystmills: Good point, I think I'll take it away for now!

.storybook/gu-preview/GuPreview.tsx
rhystmills: Good to know, I hadn't encountered that syntax but I'll give it a go.

PR #27, opened 2020-10-14
Update guardian dependencies and move to peer depencies

liywjl:
## What does this change?
Update @guardian dependencies and move them to peer dependencies to avoid bundling them in the npm package

Need to make sure Frontend and DCR have the correct matching peer dependencies

### DIFF
--- a/package.json
+++ b/package.json
@@ -70,10 +70,10 @@
     "react": "16.13.x"
   },
   "dependencies": {
-    "@guardian/src-button": "^2.1.0",
-    "@guardian/src-foundations": "^2.1.0",
-    "@guardian/src-grid": "^2.2.0",
-    "@guardian/src-icons": "^2.2.0",
+    "@guardian/src-button": "^2.4.0",
+    "@guardian/src-foundations": "^2.4.0",
+    "@guardian/src-grid": "^2.4.0",
+    "@guardian/src-icons": "^2.4.0",
     "emotion-theming": "^10.0.19"
   }
 }
--- a/yarn.lock
+++ b/yarn.lock
@@ -1256,48 +1256,36 @@
   resolved "https://registry.yarnpkg.com/@emotion/weak-memoize/-/weak-memoize-0.2.5.tgz#8eed982e2ee6f7f4e44c253e12962980791efd46"
   integrity sha512-6U71C2Wp7r5XtFtQzYrW5iKFT67OixrSxjI4MptCHzdSVlgabczzqLe0ZSgnub/5Kp4hSbpDB1tMytZY9pwxxA==
 
-"@guardian/src-button@^2.1.0":
-  version "2.1.0"
-  resolved "https://registry.yarnpkg.com/@guardian/src-button/-/src-button-2.1.0.tgz#59e4bec31dce699debcb2c49c128251e00e96ecc"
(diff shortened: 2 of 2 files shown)


package.json
```diff
@@ -29,6 +29,10 @@
     "@babel/preset-react": "^7.10.4",
     "@babel/preset-typescript": "^7.10.4",
     "@emotion/babel-preset-css-prop": "^10.0.27",
+    "@guardian/src-button": "2.4.0",
+    "@guardian/src-foundations": "2.4.0",
+    "@guardian/src-grid": "2.4.0",
+    "@guardian/src-icons": "2.4.0",
```
tjmw, line 35: Ah so these need to be duplicated here because we need them for development. I'd been experimenting with using https://www.npmjs.com/package/install-peers-cli to install peer deps for local dev. Any thoughts on one approach or the other?

tjmw (approved):
🏅 Thanks for doing this @liywjl, I think it looks good. I'd like to test against frontend & DCR before merging if that sounds okay? I'll make some time to do that today (I think the process will be build the package, `yarn link` and then force the banner to show up on the platforms).

package.json
liywjl: I think using `devDependencies` is fine, using `install-peers-cli` adds an extra layer that may not necessarily be obvious to new devs. But as long as it is clearly communicated in the ReadMe then I think its fine.

tjmw:
Hey @liywjl I did a little testing (only on DCR so far) and strangely I'm seeing a slight difference in the banner between the version that's on prod and this PR. I'm testing by forcing the banner using the query string:

[On prod](https://www.theguardian.com/sport/2019/jul/28/tour-de-france-key-moments-egan-bernal-yellow-jersey?dcr&force-braze-message=%7B%22componentName%22:%22DigitalSubscriberAppBanner%22,%22header%22:%22Well+hello+Tom%22,%22body%22:%22Something+something%22%7D):

Screenshot 2020-10-16 at 14 19 35

[Locally](http://localhost:3030/Article?url=https://m.code.dev-theguardian.com/sport/2019/jul/28/tour-de-france-key-moments-egan-bernal-yellow-jersey&force-braze-message=%7B%22componentName%22:%22DigitalSubscriberAppBanner%22,%22header%22:%22Well%20hello%20Tom%22,%22body%22:%22Something%20something%22%7D):

Screenshot 2020-10-16 at 14 19 32

Note the colour of the buttons and the x to close. I wonder if the src bump has somehow affected things. Weirdly this branch looks fine in Storybook. Any thoughts on what might have caused this?

tjmw:
Also tested on [local frontend](http://localhost:9000/sport/2019/jul/28/tour-de-france-key-moments-egan-bernal-yellow-jersey?force-braze-message=%7B%22componentName%22:%22DigitalSubscriberAppBanner%22,%22header%22:%22Well%20hello%20Tom%22,%22body%22:%22Something%20something%22%7D) and I get the following in the browser console:

Screenshot 2020-10-16 at 14 34 56

That message is a bit cryptic, but I wonder if this is because frontend isn't providing some of the dependencies which have been moved to `peerDependencies` (it looks like `src-icons` and `src-grid` aren't provided).

tjmw:
@liywjl and I chatted offline about the visual differences highlighted above (and the error I was seeing on frontend) and it looks like the move to peer dependencies was the culprit in both cases. We decided to split out the move to peer dependencies from this PR and explore that separately. We'd also like to figure out how we could have caught the visual diff automatically. I'm going to look into adding a Storybook/Chromatic test for this in DCR.

PR #44, opened 2020-11-10
New AppBanner with custom image and tertiary text

rhystmills:
## What does this change?

Introduces a new component where image and tertiary text are configurable, in order to give more agency to Marketing, allowing them to create and spin up their own campaigns.

Also includes a general refactor to help support multiple components.

The image URL prop provided must match the allowlist of hosts which currently is configured to be `['https://media.guim.co.uk/', 'https://i.guim.co.uk/']`. This is to limit the source of images which marketing can use. If the image does not match the allowlist the banner is not rendered.

This is a first step, rather than a final product, and we plan to improve how images are included in banners in future steps. This feature isn't designed to be consumed by marketing just yet, so we've hidden the new more flexible `AppBanner` in the hosted storybook instances. Follow up steps are to complete the grid picker storybook integration to give marketing a nice approach for including images. The we'd ideally like to move on to taking a selected grid image and returning a signed `i.guim.co.uk` image in storybook, which marketing can then copy across to the in-app message configuration in Braze.

## How to test

Run `yarn storybook` in your local copy of the repository.

The image integration does not account for image dimensions/size etc yet, which we will want to do.


src/AppBanner/index.stories.tsx
```diff
@@ -0,0 +1,56 @@
...
+import { StorybookWrapper } from '../utils/StorybookWrapper';
+import { knobsData } from '../utils/knobsData';
+
+export default {
+    component: 'AppBanner',
+    title: 'Components/AppBanner',
+    decorators: [withKnobs],
+    parameters: {
+        knobs: {
+            escapeHTML: false, // HTML escaped in order to prevent double-escaping of imgUrl special characters in Storybook
```
tjmw, line 13: I feel like there's a missing negative somewhere in this sentence :-) We're telling Storybook not to escape HTML as this results in double escaping/encoding by the time the URL is rendered in the browser.

src/DigitalSubscriberAppBanner/index.stories.tsx
```diff
@@ -1,8 +1,8 @@
 import React, { ReactElement } from 'react';
 import { withKnobs, text } from '@storybook/addon-knobs';
-import { BrazeMessage } from './BrazeMessage';
-import { StorybookWrapper } from './utils/StorybookWrapper';
-import { knobsData } from './utils/knobsData';
+import { BrazeMessage } from '../BrazeMessage';
```
tjmw, line 3: Reminder to ourselves that we talked about ways to avoid having to update relative paths when we move files around. I used https://github.com/tleunen/babel-plugin-module-resolver on a previous project which allows you to define aliases/named entry points and it worked well, so that could be an option (not sure how well it plays with TypeScript, we'd need to try it out).

src/AppBanner/index.stories.tsx
rhystmills: Ah yes, good spot. I've updated it now.

src/utils/images.ts
```diff
@@ -0,0 +1,4 @@
+const ALLOWED_HOSTS = ['https://media.guim.co.uk/', 'https://i.guim.co.uk/'];
+
+export const isImageUrlAllowed = (imageUrl: string): boolean =>
+    ALLOWED_HOSTS.some((host) => imageUrl.startsWith(host));
```
tjmw, line 4: I noticed that `String.prototype.startsWith` isn't supported natively on IE11, but we polyfill it (confirmed on BrowserStack).

arausuy (approved):

PR #45, opened 2020-11-10
Run Local Storybook on a gutools domain

tjmw:
## What does this change?

Add support for running Storybook locally on a gutools domain. This is so that we can embed Grid into the story for image handling (coming in a future PR).

## How to test

* Run storybook locally
* Visit https://braze-storybook.local.dev-gutools.co.uk/

## Caveats

* There are still some items to figure out which we can tackle in future PRs. We'll want to access the deployed version of Storybook via a gutools domain too (and this will need to embed prod Grid). Probably by building storybook and deploying to an S3 bucket.

* Running `yarn storybook` still opens http://localhost:6016. This is because it's non-trivial to override the URL which storybook opens in the browser. You can tell Storybook not to open in a browser with the `--ci` flag, but because the Storybook server is a daemon it's hard to know when to then `open https://braze-storybook.local.dev-gutools.co.uk/`. If you background the process the page opens before Storybook has finished building and a 503 is returned. I'm experimenting with some kind of client side redirect to make this work.

### DIFF
--- a/README.md
+++ b/README.md
@@ -5,6 +5,12 @@ frontend.
 
 ## Development
 
+### Local Setup
+
+```
+$ bin/setup
+```
+
 ### Storybook
 
 We use Storybook when building components. Run Storybook with:
--- a/.gitignore
+++ b/.gitignore
@@ -4,3 +4,4 @@ dist/
 /storybook-static/*
 build-storybook.log
 stats.html
+Brewfile.lock.json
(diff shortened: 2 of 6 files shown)


README.md
```diff
@@ -5,6 +5,12 @@ frontend.
 
 ## Development
 
+### Local Setup
+
+```
+$ bin/setup
```
adamnfish, line 11: I like the use of the `bin/*` convention here 👍

adamnfish (approved):
Great, thank you! I have tested this on my computer and it works fine 👍

tjmw:
Thanks for the review @adamnfish!

PR #49, opened 2020-11-20
Integrate Grid Picker with Storybook

tjmw:
## What does this change?

Builds on top of #44 by integrating with Grid, allowing the marketing user to select an image from Grid instead of specifying the image manually. As with the other configurable elements, the marketing user would then transfer these to Braze as key/value pairs when configuring an in-app message.

https://trello.com/c/4QSxdwus/994-integrate-grid-picker-with-storybook

Outstanding items, which could be tackled in future PRs:

* Hosted storybooks are still on Chromatic, where the grid picker won't load because of the content security policy. They need to be moved to a gutools.co.uk domain.
* We aren't being smart about which crop size to use, we simply selected the last one on the list. We've talked about extending this to generate a `i.guim.co.uk` URL from the `media.guim.co.uk` url returned by Grid. If we did that I'd think we'd use the master crop and then apply our own quality and size params to the i.guim url.
* Known issue where the currently selected image is reset when the browser window is resized below a certain width.
* Running storybook locally with `yarn storybook` opens on localhost:6016 where the grid picker won't work, again because of the CSP issue mentioned above. You have to visit https://braze-storybook.local.dev-gutools.co.uk/ manually.
* You must have already auth'd with Grid to be able to use the picker in storybook.

## Images

Screenshot 2020-11-20 at 16 44 33

### DIFF
--- a/.storybook/grid/GridPanel.tsx
+++ b/.storybook/grid/GridPanel.tsx
@@ -1,0 +1,122 @@
+import React, { useEffect, useState, useRef } from 'react';
+import { AddonPanel } from '@storybook/components';
+import { useChannel } from '@storybook/api';
+import { STORY_CHANGED } from '@storybook/core-events';
+import { API } from '@storybook/api';
+import type { IframePostMessage } from '@guardian/grid-client';
+import { styles } from './gridStyles';
+import { INITIAL_IMAGE_EVENT, IMAGE_SELECTED_EVENT } from './withGrid';
+import { GRID_URL } from '../../src/utils/env';
+
+const isValidMessage = (data: IframePostMessage) =>
+    data?.crop?.data && data?.image?.data && data?.crop?.data?.assets?.length > 0;
+
+type Props = {
+    active: boolean;
+    api: API;
+};
+
+export const GridPanel = ({ api, active }: Props) => {
+    const emit = useChannel({});
+    const [isModalOpen, setIsModalOpen] = useState<boolean>(false);
+    const [selectedImage, setSelectedImage] = useState<string | undefined>();
+    const activeButton = useRef(null);
+
+    const onMessage = (event) => {
+        if (event.origin !== GRID_URL) {
+            return;
+        }
+
+        const data: IframePostMessage = event.data;
+
+        if (!isValidMessage(data)) {
+            console.log("Image data isn't valid!", data);
+            return;
+        }
+
+        const cropAssets = data.crop.data.assets;
+        const cropUrl = cropAssets[cropAssets.length - 1]?.secureUrl;
+
+        if (!cropUrl) {
+            return null;
+        }
+        setSelectedImage(cropUrl.toString());
+        emit(IMAGE_SELECTED_EVENT, cropUrl);
+
+        setIsModalOpen(false);
+    };
+
+    useEffect(() => {
+        if (isModalOpen) {
+            window.addEventListener('message', onMessage, false);
+        } else {
+            window.removeEventListener('message', onMessage, false);
+        }
+
+        return () => window.removeEventListener('message', onMessage, false);
+    }, [isModalOpen]);
+
+    useEffect(() => {
+        // This panel is reused across stories, so reset when the story is changed
+        const onStoryChanged = () => {
+            setIsModalOpen(false);
+            setSelectedImage(undefined);
+        };
+
+        const onInitialImage = (initialImage) => {
+            setSelectedImage(initialImage);
+        };
+
+        api.on(STORY_CHANGED, onStoryChanged);
+        api.on(INITIAL_IMAGE_EVENT, onInitialImage);
+
+        return () => {
+            api.off(STORY_CHANGED, onStoryChanged);
+            api.off(INITIAL_IMAGE_EVENT, onInitialImage);
+        };
+    }, []);
+
+    return (
+        <AddonPanel active={active}>
+            <div>
+                <div css={styles.buttonRow}>
+                    <button
+                        css={isModalOpen ? styles.inactiveButton : styles.activeButton}
+                        ref={activeButton}
+                        onClick={() => {
+                            setIsModalOpen(true);
+                            activeButton.current.blur();
+                        }}
+                    >
+                        Select image
+                    </button>
+                    <button
+                        css={isModalOpen ? styles.activeButton : styles.inactiveButton}
+                        ref={activeButton}
+                        onClick={() => {
+                            setIsModalOpen(false);
+                            activeButton.current.blur();
+                        }}
+                    >
+                        Close Grid
+                    </button>
+                </div>
+
+                {selectedImage && (
+                    <div css={styles.resultRow}>
+                        <span css={styles.resultKey}>imageUrl:</span>
+                        <textarea css={styles.resultValue} value={selectedImage} readOnly={true} />
+                    </div>
+                )}
+
+                {isModalOpen && <GridModal />}
+            </div>
+        </AddonPanel>
+    );
+};
+
+const GridModal = () => (
+    <div css={styles.modal}>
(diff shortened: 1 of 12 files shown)


src/utils/images.ts
```diff
@@ -1,4 +1,10 @@
 const ALLOWED_HOSTS = ['https://media.guim.co.uk/', 'https://i.guim.co.uk/'];
+const ALLOWED_DEV_HOSTS = [
+    ...ALLOWED_HOSTS,
+    'https://s3-eu-west-1.amazonaws.com/media-origin.test.dev-guim.co.uk/',
```
paulbrown1982, line 4: We should raise a ticket with Ed Tools to get object listing removed from this bucket, especially as we have now made the bucket name public...

.storybook/grid/GridPanel.tsx
```diff
@@ -0,0 +1,115 @@
...
+        }
+
+        const postMessageService = new IframePostMessageService(event, Reporter.default);
+
+        if (!postMessageService.isValid) {
+            return;
+        }
+
+        const cropAssets = event.data.crop.data.assets;
+        const cropUrl = cropAssets[cropAssets.length - 1]?.secureUrl;
```
paulbrown1982, line 35: Won't this code break if the image has no crops (which is somewhat common)?

.storybook/grid/GridPanel.tsx
```diff
@@ -0,0 +1,115 @@
...
+
+    useEffect(() => {
+        if (isModalOpen) {
+            window.addEventListener('message', onMessage, false);
+        } else {
+            window.removeEventListener('message', onMessage, false);
+        }
+
+        return () => window.removeEventListener('message', onMessage, false);
+    }, [isModalOpen]);
```
paulbrown1982, line 50: I can't see how the inner scope `useEffect` will know the name of the `isModelOpen` variable, if it's only passed in as an array rather than an object in its second argument...

.storybook/grid/GridPanel.tsx
```diff
@@ -0,0 +1,115 @@
...
+        const onStoryChanged = () => {
+            setIsModalOpen(false);
+            setSelectedImage(undefined);
+        };
+
+        const onInitialImage = (initialImage) => {
+            setSelectedImage(initialImage);
+        };
+
+        const stopListeningToStoryChanged = api.on(STORY_CHANGED, onStoryChanged);
```
adamnfish, line 63: As a reader it isn't crystal clear to me that this "stops listening". This appears to be adding a handler that does nothing, rather than removing the existing handler? If this works then I guess this is actually mutating the existing handler reference.

No harm if this works, but if this is confusing to others as well perhaps we could think about removing the events rather than changing them to NoOps in this situation?

src/utils/images.ts
adamnfish: Great shout. We should also check that it's ok to make this bucket name public at all. We might wanna nuke the history here if not. My suspicion is that this bucket name is "private", in the three-tier "public", "private", "secret" categorisation, which means it shouldn't be in a public repo.

We're racking up use cases that need secrets, I wanna chat to you all about adding a backend to this where we can put them.

.storybook/grid/GridPanel.tsx
```diff
@@ -0,0 +1,115 @@
...
+import { useChannel } from '@storybook/api';
+import { styles } from './gridStyles';
+import { INITIAL_IMAGE_EVENT, IMAGE_SELECTED_EVENT } from './withGrid';
+import { STORY_CHANGED } from '@storybook/core-events';
+import { IframePostMessageService } from '@guardian/grid-client';
+import { Reporter } from '@guardian/grid-client/lib/utils';
+
+const GRID_URL =
+    process.env.NODE_ENV === 'development'
+        ? 'https://media.test.dev-gutools.co.uk'
```
adamnfish, line 12: I feel like these configuration constants are a bit scattered at the moment. Not urgent for now, but we might consider colocating these definitions in the future so that:

* it's easy to see what has been configured in one place
* it's simpler to keep these configurations in sync
* the around these bits of config (specifically how we detect dev mode) is not duplicated

.storybook/grid/GridPanel.tsx
adamnfish: The `isValidMessage` function checks for a crop, but doesn't seem to be used. Once you have a crop, it *should* have assets available, but maybe we want to double check with EdTools that this is true - a crop will always have at least one asset. If not, we can add this check to `isValidMessage` (once that validation is used).

.storybook/grid/GridPanel.tsx
```diff
@@ -0,0 +1,115 @@
...
+        : 'https://media.gutools.co.uk';
+
+const isValidMessage = (data) => data?.crop?.data && data?.image?.data;
```
adamnfish, line 15: Is this used?

adamnfish (commented):
Great!

This looks basically fine, but there are a couple things to look at. I do think we might want to look into the bucket string before this gets merged because life gets more difficult to fix the further that spreads. Otherwise, my other points are things we can talk about later or now, up to you both.

.storybook/grid/GridPanel.tsx
tjmw: Ah great spot, thanks, I'll remove. No it's not used any longer, `@guardian/grid-client` is now doing this for us. (I'm surprised the editor/build steps didn't shout about this, perhaps there's a linting/typescript config value we're missing to make sure it does)

.storybook/grid/GridPanel.tsx
tjmw: Agreed with all this 👍  As a first pass a single method to check whether we're in dev mode seems like a good place to start.

src/utils/images.ts
tjmw: Ahh thanks for pointing this out, that concern didn't occur to me here. Let's chat about what needs to happen with this.

.storybook/grid/GridPanel.tsx
tjmw: Hey good question. The scoping here is that the `isModelOpen` const is defined outside the first argument to `useEffect` (the callback) so it closes over it. The second (optional) argument to `useEffect` is the list of dependencies, so at the end of the current render cycle the effect will be called only if `isModalOpen` has changed.

.storybook/grid/GridPanel.tsx
tjmw: We've replaced our own homegrown `isValidMessage` with `!postMessageService.isValid` from `@guardian/grid-client` which, if I understand correctly, will return `false` if there are no crops. I'll double check that though, it'd be good to be sure. I'd say if the validation _isn't_ right it'd be good to capture those rules in `guardian/grid-client` rather than expanding on them here.

.storybook/grid/GridPanel.tsx
tjmw: Oh weirdly I'd seen examples where the return value of the `.on` function was a function to de-register the handler, but looking at the docs that's not the case. So I must have been mistaken. I'll fix. I think the more standard `api.off` is the way to go.

.storybook/grid/GridPanel.tsx
tjmw: I've done a little refactoring around this. I'm sure we can improve things further but I think I might save that for a future PR.

tjmw:
Hi @adamnfish @paulbrown1982 thanks for all the feedback, we've been through and I think we've addressed everything at this point. Regarding the allow-listing of the bucket - we decided that since we only need that for development that rather than being too specific we can loosen the validation for dev only and allow any https://s3-eu-west-1.amazonaws.com/ URL (without specifying the bucket). I think that's enough to unblock this PR - what do you reckon?

paulbrown1982 (approved):
👍  LGTM

adamnfish (approved):

PR #59, opened 2020-11-27
Update URL we use to preview the banner

tjmw:
## What does this change?

We no longer support using www so use preview instead.

See guardian/frontend#23311 and guardian/dotcom-rendering#2148.

## How to test

The little preview icon in storybook should now link to an article on preview.gutools.co.uk instead of www.theguardian.com.


rhystmills (approved):
Looks good.

PR #61, opened 2020-12-01
Deploy Built Storybook to S3

tjmw:
## What does this change?

Build and deploy Storybook to S3. This is so that we can put Storybook behind a gutools domain in order to embed Grid (added in #49) and abide by Grid's content security policy.

* Added Team City project to build storybook and upload artifacts to Riff Raff
* Cloudform the destination S3 bucket (one time only, this isn't happening on build, but I've added the cloudformation yaml to the repo)
* Configure Riff Raff to deploy to S3 using the `aws-s3` build type

### DIFF
--- a/artifact.json
+++ b/artifact.json
@@ -1,0 +1,11 @@
+{
+  "projectName": "targeting:braze-components",
+  "vcsURL": "https://github.com/guardian/braze-components",
+  "actions": [
+    {
+      "action": "braze-components-storybook-static",
+      "path": "storybook-static",
+      "compress": false
+    }
+  ]
+}
--- a/riff-raff.yaml
+++ b/riff-raff.yaml
@@ -1,0 +1,11 @@
+stacks: [targeting]
+regions: [eu-west-1]
+deployments:
+    braze-components-storybook-static:
+        app: braze-components
+        type: aws-s3
+        parameters:
+            bucket: braze-components-storybook
+            prefixStack: false
+            cacheControl: public, max-age=3600
+            publicReadAcl: false
--- a/bin/ci
+++ b/bin/ci
@@ -1,0 +1,8 @@
+#!/usr/bin/env bash
+
+set -e
+
+yarn install
+yarn test
+yarn build-storybook
+yarn upload-artifact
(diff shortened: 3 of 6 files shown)


riff-raff.yaml
```diff
@@ -0,0 +1,10 @@
+stacks: [targeting]
+regions: [eu-west-1]
+deployments:
+    storybook-static:
```
adamnfish, line 4: It occurs to me to mention that this string `storybook-static` will be used as the `App` in RiffRaff. This is quite a general name, is there something that would better fit our application? We're using `braze-components-static` elsewhere.

adamnfish: One helpful tool is the config validator in RiffRaff's UI, which shows what it understands from a RiffRaff.yaml file:

  https://riffraff.gutools.co.uk/configuration/validation

artifact.json
```diff
@@ -0,0 +1,11 @@
+{
+  "projectName": "targeting:braze-components",
+  "vcsURL": "https://github.com/guardian/braze-components",
+  "actions": [
+    {
+      "action": "storybook-static",
```
adamnfish, line 6: Don't forget to change this if we change the name in the RiffRaff.yaml file as well.

adamnfish (commented):
All looks great, although I do ask about the RiffRaff deployment name (and how that relates to the convention-derived `App`).

riff-raff.yaml
tjmw: Ah that's interesting. I'd been thinking about the app in Riff Raff as being scoped/belonging to the project. So you're right `storybook-static` is general, but scoped to `targeting:braze-components` it makes more sense. Definitely happy to change, though I'd be keen to keep `storybook` in there somewhere.

adamnfish: You can set the app explicitly in the RiffRaff.yaml if you don't want it to be drawn from the deployment name. I kind of feel like the App in RiffRaff should be the same as the App we're using elsewhere, but truthfully I'm not sure it matters heaps if its inconsistent just in RiffRaff. I feel like maybe I need to talk to DevX about whether RiffRaff's understanding of the "app" is important anywhere.

adamnfish (approved):
Great!

PR #68, opened 2020-12-08
Open Storybook on dev-gutools domain

tjmw:
## What does this change?

Running `yarn storybook` locally now correctly opens Storybook on the custom dev-gutools domain we've configured. Previously it would open on localhost:6016 and you'd have to navigate to the custom domain manually in order for the Grid integration to work.

## How to test

Run `yarn storybook` locally.

### DIFF
--- a/bin/storybook
+++ b/bin/storybook
@@ -1,0 +1,12 @@
+#!/usr/bin/env bash
+
+set -e # exit on error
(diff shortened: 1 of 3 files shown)


rhystmills (approved):
Looks good to me.

PR #69, opened 2020-12-09
Replaced NODE_ENV environment query with hostname-based solution

rhystmills:
## What does this change?

Replaced NODE_ENV environment query with hostname-based solution

Abstracts over dev, code and production environment queries.

Also surfaces isCode and isProduction methods in advance of login PR.

## How to test

(Testing in progress)


tjmw (approved):
I think this is looking good 👍 

There's one outstanding thing around the image check we we'll do in the code environment in Storybook, which will have a `NODE_ENV` of `production`, but talks to Grid test (which will return the raw s3 host for image crops). But I think we can handle that in a subsequent PR.

PR #71, opened 2020-12-13
Guardian in 2020 / Working Report Banner

tjmw:
## What does this change?

Adds a new banner (`TheGuardianIn2020Banner`) to support an end of year report. Some differences between this and the existing `DigitalSubscriberAppBanner` are:

* Different image
* Single CTA/button which is a link, rather than closing the banner
* No app store icons
* Different tertiary text

The more generic app banner we've been working on (`AppBanner`) was used as a starting point. There remains duplication between the two, though I've pulled common styles out. I'd rather wait on trying to DRY things up too much than go with the wrong abstractions at this point.

*Note:* The tracking params on the link are TBC. We'll need to update those before enabling the canvas in Braze (i.e. actually rendering this banner to users). Current plan is to merge this PR as-is so we can do UAT and then follow up to fix the tracking params.


src/TheGuardianIn2020Banner/index.tsx
```diff
@@ -0,0 +1,121 @@
...
+    try {
+        fn();
+    } catch (e) {
+        console.log(`Error (${description}): `, e.message);
+    }
+};
+
+export const COMPONENT_NAME = 'TheGuardianIn2020Banner';
+const THE_GU_IN_2020_URL =
+    'https://www.theguardian.com/info/ng-interactive/2019/dec/19/working-report';
```
tjmw, line 31: TODO: This is the URL for the previous year. Update this when we've got the new URL.

src/TheGuardianIn2020Banner/index.tsx
```diff
@@ -0,0 +1,121 @@
...
+                        <strong css={commonStyles.cta}>Hardcoded text goes here</strong>
+                    </p>
+                    <LinkButton href={THE_GU_IN_2020_URL} css={commonStyles.primaryButton}>
+                        Show me!
+                    </LinkButton>
+                </div>
+                <div css={commonStyles.bottomRightComponent}>
+                    <div css={styles.image}>
+                        <img
+                            src="https://i.guim.co.uk/img/media/de6813b4dd9b9805a2d14dd6af14ae2b48e2e19e/0_0_930_520/930.png?width=930&quality=60&s=a7d81978655765847246c8d4d0cd0e7f"
```
tjmw, line 97: TODO: Update this image URL once we have the real image.

src/TheGuardianIn2020Banner/index.tsx
adamnfish: Also reminding: We should double check it has the appropriate CMP params or other tracking on it.

src/AppBanner/styles.ts
```diff
@@ -1,163 +1,13 @@
...
-            background-color: #111;
-        }
-    `,
-
     secondaryButton: css`
         color: ${palette.neutral[20]};
     `,
 
-    packShot: css`
+    image: css`
```
adamnfish, line 10: I'm interested in why this isn't also a common style? (But no harm, I know we're deliberately duplicating stuff, which is fine)

src/TheGuardianIn2020Banner/index.tsx
```diff
@@ -0,0 +1,121 @@
...
+
+    const onCloseClick = (
+        evt: React.MouseEvent<HTMLButtonElement, MouseEvent>,
+        internalButtonId: number,
+    ): void => {
+        evt.preventDefault();
+
+        setShowBanner(false);
+
+        catchAndLogErrors('ophanButtonClick', () => {
```
adamnfish, line 49: No harm for now, but this little bit here feels like the kind of fiddly and annoying logic we'd want to write once in the future.

Ideally as a function or value we re-use from elsewhere, rather than a part of the component contract. Duplicating most of the rest looks fine to me even beyond this short-term change, means we don't end up cornering ourselves into having all the banners be the same as you say.

src/AppBanner/styles.ts
tjmw: Hey good question. Mostly because we hadn't received the real image when we did this work and I was expecting that the different shape (from the mockup) and transparency would mean the styles would have to be altered slightly for the Guardian in 2020 banner.

src/TheGuardianIn2020Banner/index.tsx
```diff
@@ -27,8 +27,25 @@ const catchAndLogErrors = (description: string, fn: () => void): void => {
...
+    utm_source: 'eml',
+    utm_medium: 'embau',
+    utm_campaign: 'MK_SU_WorkingReport2020Canvas',
+    CMP_TU: 'mrtn',
+    CMP_BUNIT: 'supporters',
+    INTCMP: 'Working_Report_2020_Retention_Banner',
+};
+
+for (const [key, value] of Object.entries(urlParams)) {
+    urlObject.searchParams.append(key, value);
```
paulbrown1982, line 45: I feel `.set` would be more suitable than `.append` as it would ensure there are no repeat parameters.

See: https://developer.mozilla.org/en-US/docs/Web/API/URLSearchParams/set

tjmw: Agree with this comment about using `set` :+1:

Purely personal preference and totally something to take or leave, but whenever I see a list of things being condensed down into a single thing like this I think [reduce](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/reduce).

rhystmills: Thanks both, have updated to use `.set()`.

I had a go using reduce, and I think the solution here would be - 
```
Object.entries(urlParams).reduce((acc, cur) => {
    const [key, value] = cur;
    acc.searchParams.set(key, value);
    return acc;
}, urlObject);
```
- which is a bit more verbose than the `for ... of` loop and I'm not sure it's more readable in this case (though perhaps my implementation is a bit clunky). Also worth noting that we're mutating the same URL object repeatedly (via `.set`) rather than actually passing forward a 'new' accumulator value each time which feels a bit misleading.

adamnfish: Agreed, mutating in a reduce is a bit misleading. We'd probably want to refactor that to be a more conventional accumulator if we were going to use reduce. OK as-is though.

src/TheGuardianIn2020Banner/index.stories.tsx
```diff
@@ -0,0 +1,48 @@
...
+
+    // This is to make the data available to the guPreview add-on:
+    knobsData.set({ header, body, componentName });
+
+    return (
+        <StorybookWrapper>
+            <BrazeMessage
+                componentName={componentName}
+                logButtonClickWithBraze={(internalButtonId) => {
+                    console.log(`Button with internal ID ${internalButtonId} clicked`);
```
adamnfish, line 36: Are we ok to leave these `console.log` lines?

adamnfish (approved):
Nice one

src/TheGuardianIn2020Banner/index.tsx
tjmw: Hey sorry I missed your update Rhys. Ah yep now you've written it out I agree with Adam, the combination of mutating and reduce isn't great. Sorry for the derailment!

src/TheGuardianIn2020Banner/index.stories.tsx
tjmw: Yeah I think this is fine. In the context of Storybook there's not much else we can do in these handlers, so we just log something so it's possibility to get visibility of them.

PR #74, opened 2020-12-18
Special Edition Banner

tjmw:
## What does this change?

Adds a new banner to support an editions app message for the end of the year. The turn around on this was tight so we've essentially duplicated `DigitalSubscriberAppBanner` changing the necessary elements.

## How to test

Storybook!

## Images

Note that the body text is likely to be made shorter (controlled by marketing in Braze):

Screenshot 2020-12-18 at 09 57 59

Screenshot 2020-12-18 at 09 58 25


rhystmills (approved):
Everything here looks right to me, great work!

tjmw:
Ha thanks @rhystmills, though I think you took care of the one complex bit (the image)! Since we both worked on this let's see if we can get another pair of 👀  on it before merging.

paulbrown1982 (approved):
👍

PR #75, opened 2020-12-18
Bump src packages

tjmw:
## What does this change?

Bumps the versions of src libs we're using. While developing another banner an issue came up with `LinkButton` which is fixed in later versions of src-button. I agreed with William that we'd bump the version user here and they'd do the same for DCR (better to have fewer versions of dependencies floating around at the same time).

## How to test

Storybook!

### DIFF
--- a/package.json
+++ b/package.json
@@ -76,10 +76,10 @@
     "react": "16.13.x"
   },
   "dependencies": {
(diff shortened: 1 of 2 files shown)


rhystmills (approved):
Looks good to me.

PR #81, opened 2021-01-12
Bump @guardian/types to v1.1.0

oliverlloyd:
## What does this change?
Bumps the `@guardian/types` lib to `v1.1.0`

**This is a breaking change**

## Why?
The move to `v1` was a breaking change  to the way these types are exported and in order for DCR to upgrade (so that we can use const enums from /types) we need all dependencies to also bump their versions.


oliverlloyd:
[@guardian/types v1.1 for braze-components](https://trello.com/c/M5YZrazc/2312-guardian-types-v11-for-braze-components)

tjmw (approved):
Looks good, thanks @oliverlloyd 👍

PR #85, opened 2021-01-12
Use np to publish

tjmw:
Other folks at The Guardian have been using [`np`](https://github.com/sindresorhus/np) to publish to NPM and it provides some [nice features](https://github.com/sindresorhus/np#why) over using plain `yarn publish`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -58,7 +58,7 @@ You'll need to be part of the @guardian NPM org. Ask someone to add you if
 you're not already.
 
 ```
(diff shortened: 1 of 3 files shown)


rhystmills (approved):
Looks great!

tjmw:
Thanks @rhystmills. Good question, the `prerelease` will run before `release` automatically (so no need to run manually). It's needed because we want to build/compile the package to `dist` before publishing.

PR #110, opened 2021-02-25
Add Braze Epic story to Storybook

rhystmills:
## What does this change?
Adds a story for the Braze Epic to Storybook.

The component itself is derived from support-dotcom-components via the contributions API.

## How to test

Run `yarn storybook`.

This story is available so that text options can be tested in the epic, but the component doesn't live in braze-components as the Braze banners do. We have provided support for up to 9 paragraphs.

The titles of our text fields (e.g. slotName) and their values provide the key value pairs to be used in Braze.

## Current issues
The GuPreview addon doesn't currently work for the Epic. Unfortunately, query strings on preview.theguardian.com cannot exceed 1024 characters, and the Epic frequently will do so.

A good solution may be using a url fragment instead of query parameters. We will need to update frontend and DCR to support this.

## Images

image


src/Epic/index.stories.tsx
```diff
@@ -0,0 +1,101 @@
...
+import type { EpicProps } from './index';
+import { knobsData } from '../utils/knobsData';
+
+export default {
+    component: 'Epic',
+    title: 'Components/Epic',
+    decorators: [withKnobs],
+    parameters: {
+        knobs: {
+            escapeHTML: false, // Block HTML escaping, preventing double-escaping of imgUrl special characters in Storybook
```
tjmw, line 14: I wonder if this is needed here, since it relates to the `imgUrl` which isn't a prop of the epic.

src/Epic/index.stories.tsx
```diff
@@ -0,0 +1,101 @@
...
+    const highlightedText = text('highlightedText', content.highlightedText);
+    const buttonText = text('buttonText', content.buttonText);
+    const buttonUrl = text('buttonUrl', content.buttonUrl);
+    const paragraphs = [];
+
+    // Add existing paragraphs as knobs, and then pad out with empty knobs
+    for (let i = 0; i < TOTAL_PARAGRAPHS; i++) {
+        const name = `paragraph${i + 1}`;
+        if (content.paragraphs[i]) {
+            paragraphs.push(text(name, content.paragraphs[i]));
```
tjmw, line 75: It took me a second to figure out what was going on here. So this works because `content.paragraphs[i]` is the default text provided further up. I wonder if it would be worth assigning `content.paragraphs[i]` to a const so we can give it a name, like `const defaultText = content.paragraphs[i]` just to make it super clear what's happening.

src/Epic/index.stories.tsx
```diff
@@ -0,0 +1,101 @@
...
+            paragraphs.push(text(name, ''));
+        }
+    }
+
+    const epicProps = buildProps({
+        heading,
+        highlightedText,
+        buttonText,
+        buttonUrl,
+        paragraphs: paragraphs.filter((p) => p != ''),
```
tjmw, line 86: Nice!

src/Epic/index.tsx
```diff
@@ -0,0 +1,65 @@
+import React, { useEffect, useState } from 'react';
```
tjmw, line 1: Though in general I think it's good to split code up into separate modules, I wonder if in this case it might be slightly misleading. I think in future we/someone else might look at this file and assume (given that it's a regular `ts` file under `src`) that it makes it into the final bundle for `@guardian/braze-components` (though I don't think it will since the only place which `import`s it is the story). Maybe inlining this in the story would remove that ambiguity, what do you reckon?

package.json
```diff
@@ -78,6 +78,7 @@
     "react": "16.13.x"
   },
   "dependencies": {
+    "@emotion/react": "^11.1.5",
```
tjmw, line 81: Oh, I'm interested in why this got added - is it needed? I don't fully know the context, but I believe there's a conversation going on with the dotcom team about maybe moving to `@emotion/react` but I don't think that's happened yet. On DCR currently the `emotion` dependency which is shared with `@guardian/braze-components` comes from `import * as emotion from 'emotion'`. Maybe we should be using that?

tjmw (commented):
Nice! This is looking great. I left a few comments, see what you reckon.

src/Epic/index.stories.tsx
rhystmills: Good point, will remove it.

src/Epic/index.stories.tsx
rhystmills: I'll do some renaming and hopefully make things clearer!

src/Epic/index.stories.tsx
```diff
@@ -0,0 +1,161 @@
...
+        slotName,
+        componentName,
+    });
+
+    // This is to make the data available to the guPreview add-on:
+    knobsData.set({ ...epicProps });
```
tjmw, line 152: Super minor point but this can probably be:

```suggestion
    knobsData.set(epicProps);
```

(I don't think we need to spread/copy `epicProps` here).

tjmw (approved):
This looks good to me, nice work!

src/Epic/index.tsx
rhystmills: I agree, they're combined now.

PR #113, opened 2021-03-06
use guardian/actions-setup-node

sndrs:
## What does this change?

uses guardian/actions-setup-node

## Why?

using NVM here installs the correct version of node, but that instance of Node is not then used for the subsequent tasks. this ensures your NVM version is the one the tasks run with.


tjmw (approved):
This is great, thanks @sndrs!

PR #115, opened 2021-03-17
Add click tracking to GuardianIn2020 banner

rhystmills:
### Context:

We had previously not included click tracking for The Guardian In 2020 banner as we believed there were Braze tracking issues related to link clicks that redirect to a new page.

However, we have tested link clicks on fast and slow internet connections and found their tracking to be reliable in both cases (click tracking in Braze does not seem to rely on asynchronous requests after the click has occurred, but instead caches the click data locally and sends it to Braze on a subsequent page view).

### What this PR adds:

As our testing has found link click tracking to be reliable, I am adding it to the GuardianIn2020 Banner. Though this banner is no longer live, it is useful to have a banner component where link click tracking is in place as an example for future work, or as a basis for a future banner.

We will also be using equivalent click tracking in the Epic component.

### DIFF
--- a/src/TheGuardianIn2020Banner/index.tsx
+++ b/src/TheGuardianIn2020Banner/index.tsx
@@ -26,6 +26,26 @@ const catchAndLogErrors = (description: string, fn: () => void): void => {
     }
 };
 
+const logToBrazeAndOphan = (internalButtonId: number): void => {
+    catchAndLogErrors('ophanButtonClick', () => {
+        // Braze displays button id from 1, but internal representation is numbered from 0
+        // This ensures that the Button ID in Braze and Ophan will be the same
+        const externalButtonId = internalButtonId + 1;
+        submitComponentEvent({
+            component: {
+                componentType: 'RETENTION_ENGAGEMENT_BANNER',
+                id: ophanComponentId,
+            },
+            action: 'CLICK',
+            value: externalButtonId.toString(10),
+        });
+    });
+
+    catchAndLogErrors('brazeButtonClick', () => {
+        logButtonClickWithBraze(internalButtonId);
+    });
+};
+
 export const COMPONENT_NAME = 'TheGuardianIn2020Banner';
 
 const urlObject = new URL(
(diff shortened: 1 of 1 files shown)


src/TheGuardianIn2020Banner/index.tsx
```diff
@@ -50,6 +50,28 @@ export const TheGuardianIn2020Banner: React.FC<Props> = ({
 }: Props) => {
     const [showBanner, setShowBanner] = useState(true);
 
+    const onClick = (internalButtonId: number): void => {
+        setShowBanner(false);
```
tjmw, line 54: Do we want to be closing the banner here?

src/TheGuardianIn2020Banner/index.tsx
```diff
@@ -50,6 +50,28 @@ export const TheGuardianIn2020Banner: React.FC<Props> = ({
...
+                    id: ophanComponentId,
+                },
+                action: 'CLICK',
+                value: externalButtonId.toString(10),
+            });
+        });
+
+        catchAndLogErrors('brazeButtonClick', () => {
+            logButtonClickWithBraze(internalButtonId);
+        });
```
tjmw, line 72: I _think_ this section is identical between the close click handler and this handler. Think we should refactor it out to a function that both can share?

tjmw (commented):
Looking good! I left a couple of questions/suggestions.

arausuy (approved):

src/TheGuardianIn2020Banner/index.tsx
rhystmills: Good question - my take on this was that we would want to close it, but I don't feel very strongly about it.

If a user were to middle click for example, they would then have to close the banner after following the link. Having both buttons close the banner makes it easier to split the audience in the next step if we were using button 0 clicks versus button 1 clicks as branched advancement criteria for a canvas (because they can't click on both)

On the other hand, a user probably wouldn't expect clicking that link to close the banner, so I wasn't certain this was the right behaviour myself. 

What do you reckon is best?

src/TheGuardianIn2020Banner/index.tsx
rhystmills: Good point - yes that makes sense.

src/TheGuardianIn2020Banner/index.tsx
tjmw: I think from a user perspective I'd expect command clicking on the link to open in a new tab and leave the banner up. That's how RR banners behave, so having consistency here would be good.

Good point about splitting the audience in the canvas, and the fact that the user could take both paths. I think as long as we let marketing know what the behaviour is then they can design the canvases accordingly.

PR #116, opened 2021-03-22
Take BrazeMessages and LocalMessageCache out of DCR 

rhystmills:
## What does this change?
Adds `BrazeMessage` and `LocalMessageCache` and their tests from DCR (currently in [this PR](https://github.com/guardian/dotcom-rendering/pull/2647)), so that they can be used on both DCR and Frontend.

This also changes the public interface by changing the name of the existing `BrazeMessage` in this repository to be `BrazeMessageComponent` because there is a naming clash with the `BrazeMessage` we are introducing from DCR. I'm open to alternative solutions, but ultimately they need to have different names as they will both be exported.

Errors are currently handled differently on DCR and Frontend for Braze Messages, so I have changed the constructor of `BrazeMessages` to take an `ErrorHandler`.
## How to test

`yarn test`


tjmw (approved):
I think this is good to merge :+1:

PR #118, opened 2021-03-23
Split Entry Points

tjmw:
## What does this change?

This adds a second entry point for the package which we'll start using in a future PR for the business logic we're adding to this package. Currently the business logic contains a single placeholder function `add(a: number, b: number): number` as a proof of concept.

## How to test

On the platforms we can now create two separate bundles, allowing us to lazy load only the code we need.

For the business logic:

```js
import(
  /* webpackChunkName: "guardian-braze-components-logic" */ '@guardian/braze-components/logic'
)
  .then((module) => {
    const add = module.add;
    // Do something with add ...
  })
```

and for the React components:

```js
import(
  /* webpackChunkName: "guardian-braze-components" */ '@guardian/braze-components'
)
  .then((module) => {
    setBrazeComponent(() => module.BrazeMessage);
  })
```

### DIFF
--- a/logic-index.ts
+++ b/logic-index.ts
@@ -1,0 +1,1 @@
+export const add = (a: number, b: number): number => a + b;
--- a/logic/package.json
+++ b/logic/package.json
@@ -1,0 +1,5 @@
+{
+  "main": "../dist/logic/index.js",
+  "module": "../dist/logic/index.esm.js",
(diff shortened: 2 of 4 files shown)


tjmw:
Thanks @rhystmills. I'll go ahead and merge this (once the build passes) and we can use this in your branch.

PR #119, opened 2021-03-24
Fix types exposed by package

tjmw:
## What does this change?

This ensures the correct TS types are available when importing /logic.


rhystmills (approved):
Looks good!

PR #120, opened 2021-03-26
Fix peer dependency issue

rhystmills:
## What does this change?

Fixes peer dependency issue that was preventing release of new package version.

## How to test

Run `yarn test` and `yarn np --no-publish --any-branch`.


tjmw (approved):
I think this looks good. The issue causing the diff Chromatic is letting us know about is fixed in another branch.

PR #122, opened 2021-03-26
Moved to new contributions epic url

rhystmills:
## What does this change?

The existing epic URL is no longer active.

This moves `braze-components` to its new URL at [https://contributions.guardianapis.com/modules/v1/epics/ContributionsEpic.js](https://contributions.guardianapis.com/modules/v1/epics/ContributionsEpic.js)

This will still need to be updated on the platforms.

## How to test

`yarn storybook`

### DIFF
--- a/src/Epic/index.stories.tsx
+++ b/src/Epic/index.stories.tsx
@@ -17,7 +17,7 @@ export default {
     },
 };
 
-const componentUrl = `https://contributions.code.dev-guardianapis.com/epic.js`;
+const componentUrl = `https://contributions.guardianapis.com/modules/v1/epics/ContributionsEpic.js`;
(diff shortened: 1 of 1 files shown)


src/Epic/index.stories.tsx
```diff
@@ -17,7 +17,7 @@ export default {
     },
 };
 
-const componentUrl = `https://contributions.code.dev-guardianapis.com/epic.js`;
+const componentUrl = `https://contributions.guardianapis.com/modules/v1/epics/ContributionsEpic.js`;
```
tjmw, line 20: 👍  Also a good move to use the prod version I think.

tjmw (approved):

PR #123, opened 2021-03-26
Emotion 11 & Source 3 Upgrade

tjmw:
## What does this change?

Upgrades the versions of Emotion and Source we're using. Following https://theguardian.design/2a1e5182b/p/54d9c6-whats-new-in-v300 and https://emotion.sh/docs/emotion-11.

## TODO

- [x] The strategy we were using to override the button colour no longer works. I've switched to using a theme, but the button colours aren't identical (see the Chromatic diff). Probably need to address this before merging.
- [ ] I'm seeing a cryptic warning (see below) when running `yarn build`. I'm not sure yet if this is causing a problem, but it'd be good to figure this out and understand the implications.

## Cryptic build error referred to above

Running `yarn build` gives some warnings:

```./index.ts → dist/index.js, dist/index.esm.js...
(!) Error when using sourcemap for reporting an error: Can't resolve original location of error.
node_modules/@guardian/src-button/dist/esm/index.js: (1:13)
(!) Error when using sourcemap for reporting an error: Can't resolve original location of error.
node_modules/@guardian/src-button/dist/esm/index.js: (1:21)
┌────────────────────────────────┐
│                                │
│   Destination: dist/index.js   │
│   Bundle Size:  50.8 KB        │
│   Minified Size:  50.77 KB     │
│   Gzipped Size:  15.14 KB      │
│                                │
└────────────────────────────────┘
┌────────────────────────────────────┐
│                                    │
│   Destination: dist/index.esm.js   │
│   Bundle Size:  50.54 KB           │
│   Minified Size:  50.51 KB         │
│   Gzipped Size:  15.12 KB          │
│                                    │
└────────────────────────────────────┘
(!) `this` has been rewritten to `undefined`
https://rollupjs.org/guide/en/#error-this-is-undefined
node_modules/@guardian/src-button/dist/esm/index.js
1: var __rest = this && this.__rest || function (s, e) {
                ^
2:   var t = {};
...and 1 other occurrence
created dist/index.js, dist/index.esm.js in 3.8s```

### DIFF
--- a/package.json
+++ b/package.json
@@ -36,7 +36,8 @@
     "@babel/preset-typescript": "^7.10.4",
     "@braze/web-sdk-core": "3.1.0",
     "@emotion/babel-preset-css-prop": "^10.0.27",
-    "@emotion/core": "10.0.x",
+    "@emotion/eslint-plugin": "^11.2.0",
+    "@emotion/react": "^11.1.2",
     "@guardian/grid-client": "^1.1.1",
     "@guardian/libs": "^1.7.0",
     "@guardian/node-riffraff-artifact": "^0.1.9",
@@ -56,12 +57,11 @@
     "@testing-library/react": "^11.0.4",
     "@types/jest": "^26.0.14",
     "@types/react": "17.0.2",
+    "@types/node": "^14.14.35",
     "@typescript-eslint/eslint-plugin": "^3.8.0",
     "@typescript-eslint/parser": "^3.8.0",
     "babel-loader": "^8.1.0",
     "chromatic": "^5.1.0",
-    "emotion": "10.0.27",
-    "emotion-theming": "^10.0.19",
     "eslint": "^7.6.0",
     "eslint-config-prettier": "^6.11.0",
     "eslint-plugin-prettier": "^3.1.4",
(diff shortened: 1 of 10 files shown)

### CODE REVIEW

File: package.json [hunk 1]
```diff
@@ -35,7 +35,8 @@
     "@babel/preset-typescript": "^7.10.4",
     "@braze/web-sdk-core": "3.1.0",
     "@emotion/babel-preset-css-prop": "^10.0.27",
-    "@emotion/core": "10.0.x",
```
tjmw, line 38: Note to my future self: check what effect this has on the Epic story where we use the window to share dependencies with the dynamically imported component. I suspect this may have broken things.

tjmw: This seems to work okay, we're still providing a dependency called `emotionCore` to the dynamically imported epic component, but it's coming from `@emotion/react` now.

liywjl:
We are nearly complete with updating to Emotion 11 on DCR https://github.com/guardian/dotcom-rendering/pull/2994

We need to finalise the work here, happy to pair to get this over the line

tsconfig.json
```diff
@@ -14,7 +14,8 @@
     "outDir": "dist",
     "sourceMap": true,
     "strictNullChecks": true,
-    "moduleResolution": "node"
+    "moduleResolution": "node",
+    "types": ["@emotion/react/types/css-prop", "node", "jest"]
```
SiAdcock, line 18: If you're using Emotion's JSX namespace, you might want to consider using [React's automatic runtime](https://reactjs.org/blog/2020/09/22/introducing-the-new-jsx-transform.html) too. This means you no longer need to write `import React from `react` at the top of every module. It also unlocks certain performance improvements. Check out Source's [recommended `tsconfig` setup](https://theguardian.design/2a1e5182b/p/24c1c7-using-emotion/t/001a69)

tjmw: Thanks @SiAdcock, good shout! We'll definitely do that, possibly in a follow up PR if that sounds reasonable? (There are a few moving parts going on with upgrading Emotion/source across DCR and other tools so am hoping not to increase the scope of this PR further).

liywjl (approved):
LGTM 👍

PR #131, opened 2021-04-06
Update React 17

liywjl:
## What does this change?

Update react to v17

## Why?

We want to support JSX

## Screenshots

Tested on the platforms by forcing the banner with `#force-braze-message`.

Frontend:

Screenshot 2021-05-13 at 11 47 16

DCR:

Screenshot 2021-05-13 at 12 13 54

### DIFF
--- a/package.json
+++ b/package.json
@@ -30,6 +30,7 @@
     "@babel/core": "^7.11.1",
     "@babel/plugin-proposal-class-properties": "^7.12.1",
     "@babel/plugin-proposal-private-methods": "^7.13.0",
+    "@babel/plugin-proposal-private-property-in-object": "^7.14.0",
     "@babel/preset-env": "^7.11.0",
     "@babel/preset-react": "^7.10.4",
     "@babel/preset-typescript": "^7.10.4",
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -70,7 +70,8 @@
     "np": "^7.2.0",
     "preact-x": "npm:preact@10.3.1",
     "prettier": "^2.0.5",
-    "react": "16.13.x",
+    "react": "17.0.2",
+    "react-dom": "^17.0.2",
```
liywjl, line 74: `react-dom` is needed in order to run unit test somehow? unsure why this is needed

tjmw: Yep I think it's needed by react-testing-library IIRC.

tjmw:
Hey @liywjl thanks for opening the PR. Sorry it's taken me a while to get to this! The React 17 upgrade is part of #123 but no sense in coupling them together so I'll plan on merging this PR soon. I checked out this branch locally and found a few steps I needed to take to get Storybook and the build working correctly. I've done a quick integration test on DCR and all seems well, I still need to do the same for frontend but I think that's the only outstanding task.

michaelclapham (approved):
Builds locally and code looks good to me!

PR #132, opened 2021-04-06
Remove Preact-X

liywjl:
## What does this change?
Remove `preact-x` dependency

## Why?
No longer used


tjmw (commented):
👍  Thanks @liywjl!

PR #142, opened 2021-04-15
Add error messages when messages expire or cannot be added to full queue

rhystmills:
## What does this change?

[Related trello card.](https://trello.com/c/9ss9sL1e/1114-actually-start-using-message-caching-on-dcr-and-frontend)

Log error messages from `LocalMessageCache` when messages expire or cannot be added to a full queue. `errorHandler` is provided by the platform (Frontend / DCR) - currently this will log to Sentry on both platforms.

The `errorHandler` must already provided to `BrazeMessages` by the platform (this was already the case).

### DIFF
--- a/src/logic/BrazeMessages.tsx
+++ b/src/logic/BrazeMessages.tsx
@@ -5,7 +5,7 @@ import { MessageCache } from './LocalMessageCache';
 import type { SlotName } from './types';
 
 type Extras = Record<string, string>;
-type ErrorHandler = (error: Error, identifier: string) => void;
+export type ErrorHandler = (error: Error, identifier: string) => void;
 
 interface BrazeMessagesInterface {
     getMessageForBanner: () => Promise<BrazeMessage>;
@@ -41,6 +41,7 @@ class BrazeMessage {
         this.slotName = slotName;
         this.cache = cache;
         this.errorHandler = errorHandler;
+        this.cache.errorHandler = errorHandler;
     }
 
     logImpression(): void {
--- a/src/logic/BrazeMessages.tsx
+++ b/src/logic/BrazeMessages.tsx
@@ -41,7 +41,6 @@ class BrazeMessage {
         this.slotName = slotName;
         this.cache = cache;
         this.errorHandler = errorHandler;
-        this.cache.errorHandler = errorHandler;
     }
 
     logImpression(): void {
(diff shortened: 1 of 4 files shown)


src/logic/LocalMessageCache.tsx
```diff
@@ -113,7 +130,7 @@ class LocalMessageCache {
     }
 
     static remove(slotName: SlotName, id: string): boolean {
-        const queue = getQueue(slotName);
+        const queue = getQueue(slotName, this.errorHandler ? this.errorHandler : undefined);
```
tjmw, line 133: I think this could be simplified to:

```js
getQueue(slotName, this.errorHandler);
```

If `errorHandler` is `undefined`, that's what we'll pass, which is what the ternary ends up doing anyway.

src/logic/BrazeMessages.tsx
```diff
@@ -41,6 +41,7 @@ class BrazeMessage {
         this.slotName = slotName;
         this.cache = cache;
         this.errorHandler = errorHandler;
+        this.cache.errorHandler = errorHandler;
```
tjmw, line 44: Totally happy to go with your judgement here but I do think setting this on the platform before passing the cache in will give us more flexibility. For example if we decided we wanted to turn off the error tracking for the cache (let's say it turns out to be really noisy) then that would be a quick change on the platform(s), whereas this way you'd need to change the package, release a new version etc.

src/logic/LocalMessageCache.tsx
```diff
@@ -79,12 +80,25 @@ const isValid = (m: CachedMessage): boolean => {
...
 };
 
-const getQueue = (slotName: SlotName): CachedMessage[] => {
+const getQueue = (slotName: SlotName, errorHandler?: ErrorHandler): CachedMessage[] => {
     const queue = readQueue(slotName);
     const validQueue = queue.filter((i) => isValid(i));
     const unexpiredQueue = validQueue.filter((i) => hasNotExpired(i));
 
     if (queue.length !== unexpiredQueue.length) {
+        const expiredMessages = queue.length - unexpiredQueue.length;
```
tjmw, line 89: Thoughts on naming this `expiredMessageCount` to make it's usage slightly clearer?

tjmw (commented):
Nice! I think this is looking good. I left a few suggestions, see what you reckon. Happy to chat over any of the comments I made if it's helpful.

What do you think about adding a few unit tests to verify that when an errorHandler is set on the cache it is invoked correctly in the right scenarios? I think you could pass a jest mock function and write assertions like `expect(dummyHandler).toHaveBeenCalledWith(...)`.

tjmw (approved):
This looks good! ✅ 

I like the solution you settled on with passing the `errorHandler` to the cache methods 👍

PR #143, opened 2021-04-16
Remove unused variable in rollup config

tjmw:
## What does this change?

We moved away from relying on dependencies via the window in #121 but deleting this `const globals` was missed.

## How to test

`yarn build` runs successfully.


rhystmills (approved):
Looks good.

PR #144, opened 2021-04-16
Add ophanComponentId knob to Braze epic story

tjmw:
## What does this change?

We realised that we need marketing to configure an `ophanComponentId` in the key value pairs sent from Braze which we'll pass with the Ophan view event. It's not passed as a prop to the component but is used on platform. Because marketing use Storybook as the staging ground for key/value pairs and then copy over to Braze, include it in the story:

Screenshot 2021-04-16 at 17 24 03

## How to test

`yarn storybook`

## Notes

In working on this we realised that the behaviour of Storybook didn't match the production behaviour on DCR where we [validate that certain props have been provided](https://github.com/guardian/dotcom-rendering/blob/acb093890c93ce6c314106bbe45e634e0be20103/src/web/lib/braze/parseBrazeEpicParams.ts#L45) (via Braze key value pairs) and render nothing if they aren't there. I've added that behaviour to Storybook but it's unfortunate that here we're duplicating the checks that we do on DCR. This should be addressed properly, but in the meantime I'm keen to have Storybook reflect the platform behaviour so this doesn't cause confusion for marketing. This isn’t an issue with the banner components as the component itself does these checks. Here we’re using the support-dotcom-components epic and aren’t in control of the prop validation. Possibly an argument for having our own epic component.

### DIFF
--- a/src/Epic/index.stories.tsx
+++ b/src/Epic/index.stories.tsx
@@ -133,6 +133,16 @@ const guPreviewOutput = (data: DataFromKnobs) => {
     );
 };
 
+const componentMappings = {
+    Epic: Epic,
+};
+
+const NullComponent = () => null;
+
+const getComponentfromName = (name: string) => {
+    return componentMappings[name] || NullComponent;
+};
+
 export const defaultStory = (): ReactElement => {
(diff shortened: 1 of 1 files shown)


rhystmills (approved):
Looks good to me!

PR #150, opened 2021-05-10
Add debug tests to launch.json

michaelclapham:
## What does this change?
Adds VSCode launch config for debugging tests

## How to test
Checkout with VSCode. Add a breakpoint to any test file. Click the Run and Debug button on the left hand side. Then choose vscode-jest-tests and click the run button next to that dropdown. This should run the tests and break at your breakpoint to allow debugging.

You can debug tests helping you fix tests faster.

We might spend too long debugging tests and go insane.

## Images
Screenshot 2021-05-10 at 13 08 24

### DIFF
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -1,0 +1,21 @@
+{
+    // Use IntelliSense to learn about possible attributes.
+    // Hover to view descriptions of existing attributes.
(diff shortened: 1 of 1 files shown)


tjmw (approved):
I've not used this before, looks useful!

PR #151, opened 2021-05-10
all method to get all messages from local message cache queue

michaelclapham:
all method to allow getting multiple valid messages from the local message cache queue.

### DIFF
--- a/src/logic/LocalMessageCache.test.ts
+++ b/src/logic/LocalMessageCache.test.ts
@@ -90,20 +90,6 @@ describe('LocalMessageCache', () => {
             expect(messages.length).toEqual(0);
         });
 
-        it('returns the first unexpired message', () => {
-            const message1 = JSON.parse(message1Json);
-            const message2 = JSON.parse(message2Json);
-            const queue = [
-                buildExpiredMessage(message1, '1'),
-                buildUnexpiredMessage(message2, '2'),
-            ];
-            setQueue('EndOfArticle', queue);
-
-            const messages = LocalMessageCache.all('EndOfArticle', appboy, noopErrorHandler);
-
-            expect(messages[0].message).toEqual(hydrateMessage(message2, appboy));
-        });
-
         it('returns the only unexpired messages', () => {
             const message1 = JSON.parse(message1Json);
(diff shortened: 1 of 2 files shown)


src/logic/LocalMessageCache.tsx
```diff
@@ -132,6 +137,23 @@ class LocalMessageCache {
...
     }
 
+    static all(
+        slotName: SlotName,
+        appboyInstance: typeof appboy,
+        errorHandler: ErrorHandler,
+    ): MessageWithId[] {
+        const queue = getQueue(slotName, errorHandler);
+
+        if (queue) {
```
tjmw, line 147: No harm, but looking at the way `peek` uses `getQueue` and the return type of `getQueue` I'm not sure this condition is necessary. It looks like we'll always get an array of type `CachedMessage[]` back so we can probably always just return `queue.map(...)` here. What do you reckon?

tjmw (commented):
Nice! This is looking good. I left a couple of minor questions/suggestions, see what you think. Happy to chat over if anything isn't clear.

src/logic/LocalMessageCache.test.ts
```diff
@@ -50,6 +50,140 @@ const buildExpiredMessage = (message: Message, id: string): CachedMessage => ({
...
+
+        it('returns empty list if the queue is empty', () => {
+            setQueue('EndOfArticle', []);
+
+            const messages = LocalMessageCache.all('EndOfArticle', appboy, noopErrorHandler);
+
+            expect(messages.length).toEqual(0);
+        });
+
+        it('returns the first unexpired message', () => {
```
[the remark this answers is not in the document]
michaelclapham: Fixed in https://github.com/guardian/braze-components/pull/151/commits/5527dd89c3da9a5fddfc48473e1892678d6132b2

src/logic/LocalMessageCache.tsx
michaelclapham: Fixed in https://github.com/guardian/braze-components/pull/151/commits/ec8a83027ff6992bcef55ef2b9f874c5a3b02e26

tjmw (approved):
💯

PR #153, opened 2021-05-11
Use pkgu to compile typescript

michaelclapham:
Use https://github.com/guardian/pkgu to compile typescript.

This simplifies our build process and allows treeshaking by consuming projects (dotcom-rendering) without split entry points.


tjmw:
I'm going to close this for now - I don't think our multiple entry points approach is compatible with pkgu. I'll keep the branch around, so we can always revisit in future.

PR #155, opened 2021-05-13
BrazeMessages returns first valid message in cache

michaelclapham:
We don't render components from invalid message on the top of cache, and instead find first message in the cache that can be rendered.

﻿- Wip: experimenting with canRender for components
- BrazeMessageComponent takes a list of candidates
- BrazeMessageComponent top level canRender
- App banner can render logic moved to canRender
- Can render logic for other components
- Updated storybook to use new candidates prop
- Braze messages filters non render-able components
- Use pkgu
- Remove canRender from components
- Move can render into seperate files
- Switch back to rollup
- Remove candidates prop from tests
- Fix braze messages tests
- Revert package.json to main version

### DIFF
--- a/logic/package.json
+++ b/logic/package.json
@@ -1,5 +1,6 @@
 {
   "main": "../dist/logic/index.js",
   "module": "../dist/logic/index.esm.js",
-  "types": "../dist/logic-index.d.ts"
+  "types": "../dist/logic-index.d.ts",
+  "sideEffects": false
 }
--- a/src/AppBanner/canRender.ts
+++ b/src/AppBanner/canRender.ts
@@ -1,0 +1,16 @@
+import type { BrazeMessageProps } from './index';
+import { isImageUrlAllowed } from '../utils/images';
+
+export const COMPONENT_NAME = 'AppBanner';
+
+export const canRender = (brazeMessageProps: BrazeMessageProps): boolean => {
+    const { header, body, cta, imageUrl } = brazeMessageProps;
+    if (!header || !body || !cta || !imageUrl) {
+        return false;
+    }
+    if (!isImageUrlAllowed(imageUrl)) {
+        console.log(`Image URL ${imageUrl} is not allowed`);
+        return false;
+    }
+    return true;
+};
--- a/src/AppBanner/index.stories.tsx
+++ b/src/AppBanner/index.stories.tsx
@@ -39,17 +39,13 @@ export const defaultStory = (): ReactElement => {
     return (
         <StorybookWrapper>
             <BrazeMessageComponent
-                candidates={[
-                    {
-                        componentName: componentName,
-                        brazeMessageProps: {
-                            header,
-                            body,
-                            imageUrl,
-                            cta,
-                        },
-                    },
-                ]}
+                componentName={componentName}
+                brazeMessageProps={{
+                    header,
+                    body,
+                    imageUrl,
+                    cta,
+                }}
                 logButtonClickWithBraze={(internalButtonId) => {
                     console.log(`Button with internal ID ${internalButtonId} clicked`);
                 }}
--- a/src/AppBanner/index.stories.tsx
+++ b/src/AppBanner/index.stories.tsx
@@ -39,18 +39,22 @@ export const defaultStory = (): ReactElement => {
     return (
         <StorybookWrapper>
             <BrazeMessageComponent
-                componentName={componentName}
+                candidates={[
+                    {
+                        componentName: componentName,
+                        brazeMessageProps: {
+                            header,
+                            body,
+                            imageUrl,
+                            cta,
+                        },
+                    },
+                ]}
                 logButtonClickWithBraze={(internalButtonId) => {
                     console.log(`Button with internal ID ${internalButtonId} clicked`);
                 }}
                 submitComponentEvent={(componentEvent) => {
                     console.log('submitComponentEvent called with: ', componentEvent);
-                }}
-                brazeMessageProps={{
-                    header,
-                    body,
-                    imageUrl,
-                    cta,
                 }}
             />
         </StorybookWrapper>
--- a/src/AppBanner/index.test.tsx
+++ b/src/AppBanner/index.test.tsx
@@ -90,4 +90,49 @@ describe('AppBanner', () => {
         // Nothing rendered as the imageUrl was not allowed
         expect(container.firstChild).toBeNull();
     });
+
+    it('does not render the banner when no body text provided', () => {
+        const { brazeMessageProps, ...props } = baseProps();
+        const invalidMessageProps = {
+            ...brazeMessageProps,
+            body: undefined,
+        };
+
+        const { container } = render(
+            <AppBanner {...props} brazeMessageProps={invalidMessageProps} />,
+        );
+
+        // Nothing rendered as the imageUrl was not allowed
+        expect(container.firstChild).toBeNull();
+    });
+
+    it('does not render the banner when no cta (call to action) text provided', () => {
+        const { brazeMessageProps, ...props } = baseProps();
+        const invalidMessageProps = {
+            ...brazeMessageProps,
+            cta: undefined,
+        };
+
+        const { container } = render(
+            <AppBanner {...props} brazeMessageProps={invalidMessageProps} />,
+        );
+
+        // Nothing rendered as the imageUrl was not allowed
+        expect(container.firstChild).toBeNull();
+    });
+
+    it('does not render the banner when no header text provided', () => {
+        const { brazeMessageProps, ...props } = baseProps();
+        const invalidMessageProps = {
+            ...brazeMessageProps,
+            header: undefined,
+        };
+
+        const { container } = render(
+            <AppBanner {...props} brazeMessageProps={invalidMessageProps} />,
+        );
+
+        // Nothing rendered as the imageUrl was not allowed
+        expect(container.firstChild).toBeNull();
+    });
 });
--- a/src/AppBanner/index.tsx
+++ b/src/AppBanner/index.tsx
@@ -34,17 +34,36 @@ const catchAndLogErrors = (description: string, fn: () => void): void => {
 export const COMPONENT_NAME = 'AppBanner';
 
 const canRender = (props: Props) => {
-    console.log(props);
+    const {
+        brazeMessageProps: { header, body, cta, imageUrl },
+    } = props;
+    if (!header || !body || !cta || !imageUrl) {
+        return false;
+    }
+    if (!isImageUrlAllowed(imageUrl)) {
+        console.log(`Image URL ${imageUrl} is not allowed`);
+        return false;
+    }
     return true;
 };
 
-const AppBanner = ({
-    logButtonClickWithBraze,
-    submitComponentEvent,
-    ophanComponentId = COMPONENT_NAME,
-    brazeMessageProps: { header, body, cta, imageUrl },
-}: Props): ReactElement | null => {
+const AppBanner = (props: Props): ReactElement | null => {
+    const {
+        logButtonClickWithBraze,
+        submitComponentEvent,
+        ophanComponentId = COMPONENT_NAME,
+        brazeMessageProps: { header, body, cta, imageUrl },
+    } = props;
+
+    if (!canRender(props)) {
+        return null;
+    }
+
     const [showBanner, setShowBanner] = useState(true);
+
+    if (!showBanner) {
+        return null;
+    }
 
     const onCloseClick = (
         evt: React.MouseEvent<HTMLButtonElement, MouseEvent>,
@@ -73,14 +92,10 @@ const AppBanner = ({
         });
     };
 
-    if (!showBanner || !header || !body || !cta || !imageUrl) {
-        return null;
-    }
-
-    if (!isImageUrlAllowed(imageUrl)) {
-        console.log(`Image URL ${imageUrl} is not allowed`);
-        return null;
-    }
+    // if (!isImageUrlAllowed(imageUrl)) {
+    //     console.log(`Image URL ${imageUrl} is not allowed`);
+    //     return null;
+    // }
 
     return (
         <div css={commonStyles.wrapper}>
--- a/src/AppBanner/index.tsx
+++ b/src/AppBanner/index.tsx
@@ -11,16 +11,18 @@ import { styles as commonStyles } from '../styles/bannerCommon';
 import { styles } from './styles';
 import { isImageUrlAllowed } from '../utils/images';
 
+type BrazeMessageProps = {
+    header?: string;
+    body?: string;
+    cta?: string;
+    imageUrl?: string;
+};
+
 export type Props = {
     logButtonClickWithBraze: BrazeClickHandler;
     submitComponentEvent: (componentEvent: OphanComponentEvent) => void;
     ophanComponentId?: string;
-    brazeMessageProps: {
-        header?: string;
-        body?: string;
-        cta?: string;
-        imageUrl?: string;
-    };
+    brazeMessageProps: BrazeMessageProps;
 };
 
 const catchAndLogErrors = (description: string, fn: () => void): void => {
@@ -33,10 +35,8 @@ const catchAndLogErrors = (description: string, fn: () => void): void => {
 
 export const COMPONENT_NAME = 'AppBanner';
 
-const canRender = (props: Props) => {
-    const {
-        brazeMessageProps: { header, body, cta, imageUrl },
-    } = props;
+const canRender = (brazeMessageProps: BrazeMessageProps): boolean => {
+    const { header, body, cta, imageUrl } = brazeMessageProps;
     if (!header || !body || !cta || !imageUrl) {
         return false;
     }
@@ -55,7 +55,7 @@ const AppBanner = (props: Props): ReactElement | null => {
         brazeMessageProps: { header, body, cta, imageUrl },
     } = props;
 
-    if (!canRender(props)) {
+    if (!canRender(props.brazeMessageProps)) {
         return null;
     }
 
@@ -158,4 +158,4 @@ const AppBanner = (props: Props): ReactElement | null => {
 
 AppBanner.canRender = canRender;
 
-export { AppBanner };
+export { AppBanner, canRender };
--- a/src/AppBanner/index.tsx
+++ b/src/AppBanner/index.tsx
@@ -9,9 +9,11 @@ import { PlayStore } from '../assets/play-store';
 import { BrazeClickHandler } from '../utils/tracking';
 import { styles as commonStyles } from '../styles/bannerCommon';
 import { styles } from './styles';
-import { isImageUrlAllowed } from '../utils/images';
 
-type BrazeMessageProps = {
+import { canRender, COMPONENT_NAME } from './canRender';
+export { COMPONENT_NAME };
+
+export type BrazeMessageProps = {
     header?: string;
     body?: string;
     cta?: string;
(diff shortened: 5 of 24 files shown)


tjmw:
@michaelclapham not sure if you noticed but the UI Tests build check is failing. It looks like the Guardian in 2020 banner has stopped rendering, possibly a sign we've got something wrong in the `canRender` for that banner?

logic/package.json
```diff
@@ -1,5 +1,6 @@
 {
   "main": "../dist/logic/index.js",
   "module": "../dist/logic/index.esm.js",
-  "types": "../dist/logic-index.d.ts"
+  "types": "../dist/logic-index.d.ts",
+  "sideEffects": false
```
tjmw, line 5: Is this a left over artifact from our experiment with pkgu, or do we need it on this branch?

src/AppBanner/canRender.ts
```diff
@@ -0,0 +1,16 @@
+import type { BrazeMessageProps } from './index';
+import { isImageUrlAllowed } from '../utils/images';
+
+export const COMPONENT_NAME = 'AppBanner';
+
+export const canRender = (brazeMessageProps: BrazeMessageProps): boolean => {
```
tjmw, line 6: I'm not sure where the right place is, but it feels like a comment explaining why this is in a separate file would be useful.

src/AppBanner/index.stories.tsx
```diff
@@ -40,18 +40,18 @@ export const defaultStory = (): ReactElement => {
...
                     body,
                     imageUrl,
                     cta,
                 }}
+                logButtonClickWithBraze={(internalButtonId) => {
+                    console.log(`Button with internal ID ${internalButtonId} clicked`);
+                }}
+                submitComponentEvent={(componentEvent) => {
+                    console.log('submitComponentEvent called with: ', componentEvent);
+                }}
```
tjmw, line 54: I guess this moved around as we experimented with the interface to `BrazeMessageComponent`. Might be worth moving back to reduce the size of the diff on this PR? What do you reckon?

src/AppBanner/index.test.tsx
```diff
@@ -90,4 +90,49 @@ describe('AppBanner', () => {
...
+        const invalidMessageProps = {
+            ...brazeMessageProps,
+            body: undefined,
+        };
+
+        const { container } = render(
+            <AppBanner {...props} brazeMessageProps={invalidMessageProps} />,
+        );
+
+        // Nothing rendered as the imageUrl was not allowed
```
tjmw, line 105: I don't think this comment is accurate 😃  - probably copy pasted from another test (also exists in other tests below)

src/AppBanner/index.tsx
```diff
@@ -68,14 +80,10 @@ export const AppBanner: React.FC<Props> = ({
...
-    }
-
-    if (!isImageUrlAllowed(imageUrl)) {
-        console.log(`Image URL ${imageUrl} is not allowed`);
-        return null;
-    }
+    // if (!isImageUrlAllowed(imageUrl)) {
+    //     console.log(`Image URL ${imageUrl} is not allowed`);
+    //     return null;
+    // }
```
tjmw, line 86: I think this commented check can be removed (now happens in the `canRender` function).

src/AppBanner/index.tsx
```diff
@@ -135,3 +143,5 @@ export const AppBanner: React.FC<Props> = ({
         </div>
     );
 };
+
+export { AppBanner };
```
tjmw, line 147: No strong preference either way but just noting that we could go back to the inline `export const AppBanner` if we wanted. What do you reckon?

src/BrazeMessageComponent.tsx
```diff
@@ -25,8 +26,10 @@ type CommonComponentProps = {
     brazeMessageProps: BrazeMessageProps;
 };
 
+export type BrazeComponent<P> = React.FC<P>;
```
tjmw, line 29: This type doesn't feel like it's pulling its weight any longer! Since we're not extending it any more. What do you think about losing it?

src/DigitalSubscriberAppBanner/canRender.ts
```diff
@@ -0,0 +1,10 @@
+import type { BrazeMessageProps } from './index';
+export const COMPONENT_NAME = 'DigitalSubscriberAppBanner';
+
+export const canRender = (brazeMessageProps: BrazeMessageProps): boolean => {
+    const { header, body } = brazeMessageProps;
+    if (!header || !body) {
+        return false;
+    }
+    return true;
```
tjmw, line 9: No strong preference, but we could write this as:

```suggestion
    return Boolean(header && body);
```

src/DigitalSubscriberAppBanner/index.stories.tsx
```diff
@@ -28,16 +28,16 @@ export const defaultStory = (): ReactElement => {
...
                     logButtonClickWithBraze={(internalButtonId) => {
                         console.log(`Button with internal ID ${internalButtonId} clicked`);
                     }}
                     submitComponentEvent={(componentEvent) => {
                         console.log('submitComponentEvent called with: ', componentEvent);
                     }}
-                    brazeMessageProps={{
-                        header: header,
-                        body: body,
-                    }}
```
tjmw, line 40: Same comment as above about moving this back just to reduce the size of the diff. No big deal obviously!

src/DigitalSubscriberAppBanner/index.tsx
```diff
@@ -2,22 +2,26 @@ import React from 'react';
...
+    brazeMessageProps: BrazeMessageProps;
 };
 
-export const COMPONENT_NAME = 'DigitalSubscriberAppBanner';
 const cta = 'Search for "Guardian live news"';
 const imageUrl =
     'https://i.guim.co.uk/img/media/de6813b4dd9b9805a2d14dd6af14ae2b48e2e19e/0_0_930_520/930.png?width=930&quality=60&s=a7d81978655765847246c8d4d0cd0e7f';
 
-export const DigitalSubscriberAppBanner: React.FC<Props> = ({
+const DigitalSubscriberAppBanner: BrazeComponent<Props> = ({
```
tjmw, line 24: Is it weird that `DigitalSubscriberAppBanner` doesn't call its own `canRender`? Under the hood we'll delegate to the `AppBanner` `canRender` but I feel like it could be confusing in future to add to the `DigitalSubscriberAppBanner` `canRender` and that wouldn't have an effect on the rendering. What do you think?

src/SpecialEditionBanner/index.tsx
```diff
@@ -2,22 +2,26 @@ import React from 'react';
 import { AppBanner } from '../AppBanner';
 import type { OphanComponentEvent } from '@guardian/types';
 import type { BrazeClickHandler } from '../utils/tracking';
+import { BrazeComponent } from '../BrazeMessageComponent';
+import { COMPONENT_NAME } from './canRender';
+export { COMPONENT_NAME };
```
tjmw, line 7: If we need to export this, shall we combine with the export at the bottom of the file?

src/TheGuardianIn2020Banner/canRender.ts
```diff
@@ -0,0 +1,8 @@
+import type { BrazeMessageProps } from './index';
+
+export const COMPONENT_NAME = 'TheGuardianIn2020Banner';
+
+export const canRender = (brazeMessageProps: BrazeMessageProps): boolean => {
+    const { header, body } = brazeMessageProps;
+    return Boolean(header && body);
```
tjmw, line 7: Oh ha I made the same suggestion to rewrite to this above. I promise I didn't see this first :-)

src/TheGuardianIn2020Banner/index.stories.tsx
```diff
@@ -31,17 +31,18 @@ export const defaultStory = (): ReactElement => {
     return (
         <StorybookWrapper>
             <BrazeMessageComponent
-                componentName={componentName}
+                candidates={[
```
tjmw, line 34: I don't think this is the interface to `BrazeMessageComponent` any more right?

src/logic/BrazeMessages.test.ts
```diff
@@ -59,7 +61,12 @@ describe('BrazeMessages', () => {
 
                 const bannerMessage = buildMessage({
                     ...JSON.parse(message1Json),
-                    extras: { slotName: 'Banner', title: 'Example' },
+                    extras: {
+                        slotName: 'Banner',
+                        componentName: G2020_BANNER_NAME,
+                        header: 'Header',
+                        body: 'Body',
+                    },
```
tjmw, line 69: Thoughts on extracting these "good" extras out to a named `const`?

src/logic/BrazeMessages.tsx
```diff
@@ -3,15 +3,51 @@
...
+    string,
+    (brazeMessageProps: Record<string, unknown>) => boolean
+> = {
+    [APP_BANNER_NAME]: appBannerCanRender,
+    [DIGITAL_SUBSCRIBER_APP_BANNER]: digitialSubCanRender,
+    [SPECIAL_EDITION_BANNER]: specialEdCanRender,
+    [G_IN_2020_BANNER_NAME]: gIn2020BannerCanRender,
+};
+
+export const canRenderBrazeMsg = (msgExtras: Extras | undefined): boolean => {
```
tjmw, line 41: I wonder if this function and the mappings etc should sit with `BrazeMessageComponent`? It feels somewhat analogous to the `canRender` function which sits with each component, but higher level. Maybe in `src/BrazeMessageComponent/canRender`?

src/logic/BrazeMessages.tsx
```diff
@@ -123,13 +159,17 @@ class BrazeMessages implements BrazeMessagesInterface {
 
     getMessageForBanner(): Promise<BrazeMessage> {
         // If there's already a message in the cache, return it
-        const messageFromCache = this.cache.peek('Banner', this.appboy, this.errorHandler);
+        const messagesFromCache = this.cache.all('Banner', this.appboy, this.errorHandler);
 
-        if (messageFromCache) {
+        const [firstValidMessage] = messagesFromCache.filter((msg) =>
```
tjmw, line 164: What do you think of the name `firstRenderableMessage`? Just to be really clear what we mean here.

src/logic/BrazeMessages.tsx
```diff
@@ -123,13 +159,17 @@ class BrazeMessages implements BrazeMessagesInterface {
 
     getMessageForBanner(): Promise<BrazeMessage> {
         // If there's already a message in the cache, return it
-        const messageFromCache = this.cache.peek('Banner', this.appboy, this.errorHandler);
```
tjmw, line 126: Just noting that we've only made these changes to `getMessageForBanner` but not `getMessageForEndOfArticle`. That's totally fine if it's what we meant to do, but wanted to mention. What do you reckon, should we change both?

src/logic/test.json
```diff
@@ -0,0 +1,109 @@
+{
```
tjmw, line 1: Not sure this should have been committed!

src/logic/LocalMessageCache.tsx
```diff
@@ -132,6 +137,23 @@ class LocalMessageCache {
         return;
     }
 
+    static all(
```
tjmw, line 140: I guess if we merge #151 first, then we can rebase this PR and this will be removed from the diff?

tsconfig.json
```diff
@@ -1,22 +1,21 @@
 {
   "compilerOptions": {
-    "target": "ES2020",
```
tjmw, line 3: Again, not sure if these changes belong in this PR? Probably left over from out pkgu experiment, which is now in a separate branch.

tjmw (commented):
This is looking good! I left a bunch of comments/questions but super happy to work through it all together!

src/DigitalSubscriberAppBanner/canRender.ts
michaelclapham: Done!

src/logic/BrazeMessages.tsx
michaelclapham: Done!

logic/package.json
michaelclapham: Removed it and the logic bundle size is still at 7kb and the DCR bundle didn't increase in size, so removing.

src/AppBanner/index.stories.tsx
michaelclapham: Done

src/AppBanner/canRender.ts
michaelclapham: Added a comment to all canRender files

michaelclapham:
@tjmw Merged latest main into this branch and this is ready for re-review

tjmw:
Super, thanks @michaelclapham I'll take a look this afternoon! It looks like the Guardian in 2020 banner is failing to render (that's why the UI tests build check is still pending), probably related to the `canRender` I expect?

src/AppBanner/canRender.ts
```diff
@@ -0,0 +1,19 @@
+import type { BrazeMessageProps } from './index';
+import { isImageUrlAllowed } from '../utils/images';
+
+/** These are in a seperate file to enable tree shaking of the logic deciding if a Braze message can be rendered
+ * this means the user won't download the Braze components bundle when the component can't be shown.
+ */
```
tjmw, line 6: 👍

src/Epic/canRender.ts
```diff
@@ -0,0 +1,26 @@
+import { BrazeMessageProps } from './index.stories';
+
+/** These are in a seperate file to enable tree shaking of the logic deciding if a Braze message can be rendered
+ * this means the user won't download the Braze components bundle when the component can't be shown.
+ */
+export const canRenderEpic = (brazeMessageProps: BrazeMessageProps): boolean => {
```
tjmw, line 6: I wonder if it's worth migrating over the tests from DCR for this too, what do you think?

src/Epic/canRender.ts
```diff
@@ -0,0 +1,26 @@
+import { BrazeMessageProps } from './index.stories';
```
tjmw, line 1: It seems _slightly_ odd to be importing this from a story file. I guess it won't end up in the prod bundle because it's a type, but I think I'd vote to move `BrazeMessageProps` to its own file and import it here (and in the story if we need it there too).

src/TheGuardianIn2020Banner/index.stories.tsx
tjmw: Oh, I think this is what's causing the diff in Chromatic. It looks like this story rendering correctly because we're passing the props in the wrong way here.

src/logic/test.json
tjmw: Bump! 😄

tsconfig.json
tjmw: Bump! 😄

src/logic/BrazeMessages.tsx
```diff
@@ -178,26 +153,26 @@ class BrazeMessages implements BrazeMessagesInterface {
 
         // Otherwise we'll wait for a fresh message to arrive, returning the
         // data from the cache (where it will have already been added)
-        return this.freshEndOfArticleMessage.then(() => {
-            const freshMessageFromCache = this.cache.peek(
-                'EndOfArticle',
-                this.appboy,
-                this.errorHandler,
+        return this.freshBannerMessage.then(() => {
```
tjmw, line 156: I think because we're in the generic `getMessageForSlot` method here that this should also be generic (either `freshBannerMessage` or `freshEndOfArticleMessage` depending on the `slotName`). I'm surprised that this didn't cause a failing test. Would be good to fix this by writing a suitable failing test first so we've got that confidence going forward.

src/logic/BrazeMessages.tsx
```diff
@@ -122,54 +123,28 @@ class BrazeMessages implements BrazeMessagesInterface {
...
-                    'Banner',
-                    this.cache,
-                    this.errorHandler,
-                );
-            }
-
-            // Generally we don't expect to reach this point
-            throw new Error('No messages for Banner slot');
-        });
+        return this.getMessageForSlot('Banner');
```
tjmw, line 126: Great!

tjmw (commented):
Hey I took another look over the changes. Great work sticking with this! I left a few more comments and bumped a couple of items from before which I think need addressing. Please shout if you want to pair on what's left, or if you have any questions!

src/logic/test.json
michaelclapham: Removed!

src/logic/BrazeMessages.tsx
```diff
@@ -78,9 +79,7 @@ class BrazeMessage {
 class BrazeMessages implements BrazeMessagesInterface {
     appboy: typeof appboy;
 
-    freshBannerMessage: Promise<appboy.InAppMessage>;
-
-    freshEndOfArticleMessage: Promise<appboy.InAppMessage>;
+    freshMessageBySlot: Record<SlotName, Promise<appboy.InAppMessage>>;
```
tjmw, line 82: Nice, this looks great 👍

src/Epic/canRender.ts
tjmw: I've added tests for this. I adapted the tests from DCR though they're slightly different. On DCR we're doing a mix of validation and translation of the props from the format we get from Braze to the format the Epic component expects in one step. It's (yet another) sign that borrowing the Epic component is a bit odd. I think this stuff would feel more natural if we brought the Epic component into this repo. That way we'd own the interface and wouldn't need to do this translation step from the Braze data.

michaelclapham: Agreed, but that'd be for another PR :)

src/logic/LocalMessageCache.tsx
michaelclapham: Hope so!

tsconfig.json
michaelclapham: Reverted to keep tsconfig.json same as on main

src/SpecialEditionBanner/index.tsx
michaelclapham: Given we're no longer exporting the component at the bottom of the file, keeping this export next to the import I think is easier to follow.

src/logic/BrazeMessages.tsx
michaelclapham: This is fixed now :)

src/logic/BrazeMessages.tsx
```diff
@@ -178,26 +153,26 @@ class BrazeMessages implements BrazeMessagesInterface {
...
-                    'EndOfArticle',
+                    slotName,
                     this.cache,
                     this.errorHandler,
                 );
             }
 
             // Generally we don't expect to reach this point
-            throw new Error('No messages for EndOfArticle slot');
+            throw new Error(`No messages for ${slotName} slot`);
```
adamnfish, line 175: Would it be helpful to say no **valid** messages for the slot, here?

adamnfish (approved):
Might be good to run through this with you both on a call, but as far as I can see this looks great!

src/Epic/canRender.test.ts
```diff
@@ -0,0 +1,140 @@
...
+            buttonText: 'Button',
+            buttonUrl: 'https://www.example.com',
+        };
+
+        const got = canRenderEpic(dataFromBraze);
+
+        expect(got).toEqual(false);
+    });
+
+    it('returns the correct error message when multiple fields missing', () => {
```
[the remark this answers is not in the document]
tjmw: Good spot, thanks!

src/AppBanner/index.tsx
```diff
@@ -32,16 +37,24 @@ const catchAndLogErrors = (description: string, fn: () => void): void => {
     }
 };
 
-export const COMPONENT_NAME = 'AppBanner';
+export const AppBanner = (props: Props): ReactElement | null => {
+    const {
+        logButtonClickWithBraze,
+        submitComponentEvent,
+        ophanComponentId = COMPONENT_NAME,
+        brazeMessageProps: { header, body, cta, imageUrl },
```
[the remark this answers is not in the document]
tjmw: Ah the type _is_ `BrazeMessageProps` (see the `Props` type definition on line 25, and then our use of it on line 40 in the signature for this component). This line is destructuring the `brazeMessageProps` property into `const`s called `header`, `body` etc.

PR #157, opened 2021-05-21
Add TypeScript build check

tjmw:
## What does this change?

Add TypeScript build check.

## How to test

A new build check in the GitHub UI - `typescript / typescript (push)`.

### DIFF
--- a/.github/workflows/typescript.yml
+++ b/.github/workflows/typescript.yml
@@ -1,0 +1,18 @@
+name: typescript
+on: push
+
+jobs:
+    typescript:
+        runs-on: ubuntu-latest
+        steps:
+            - uses: actions/checkout@v2
+            - uses: guardian/actions-setup-node@main
+
+            # Cache npm dependencies using https://github.com/bahmutov/npm-install
+            - uses: bahmutov/npm-install@v1
+
+            - name: Install
+              run: yarn
+
+            - name: Check typescript
+              run: yarn tsc


.github/workflows/typescript.yml
```diff
@@ -0,0 +1,18 @@
...
+
+jobs:
+    typescript:
+        runs-on: ubuntu-latest
+        steps:
+            - uses: actions/checkout@v2
+            - uses: guardian/actions-setup-node@main
+
+            # Cache npm dependencies using https://github.com/bahmutov/npm-install
+            - uses: bahmutov/npm-install@v1
```
michaelclapham, line 12: Brilliant action! Hopefully will save a lot of build time

michaelclapham (approved):
It builds! Action looks good to me

.github/workflows/typescript.yml
```diff
@@ -0,0 +1,18 @@
...
+on: push
+
+jobs:
+    typescript:
+        runs-on: ubuntu-latest
+        steps:
+            - uses: actions/checkout@v2
+            - uses: guardian/actions-setup-node@main
+
+            # Cache npm dependencies using https://github.com/bahmutov/npm-install
```
adamnfish, line 11: This looks like a useful action!

Out of interest (and not worth stressing about here), do we as a department have guidelines about when we use marketplace actions and when we should stick to the built in ones (with a little more config)?

This is achievable with the GitHub cache action, which is:

## a lot more flexible and powerful

This means we'd be able to use the same tool for lots of workflows, rather than having to use different bespoke ones each time.

## supported by GitHub

This is useful in both operational and security terms. It'll certainly keep working, and it'll certainly be kept up to date with security patches. (It's also much more likely to have its source code stay secure)

## harder to use

Obviously for this workflow the bahmutov action is loads simpler, so that's a huge plus. I wonder if the actions folks want to set some guidelines that will help teams balance these concerns?

adamnfish (commented):
I'm interested to hear why is this in a separate workflow, instead of being part of the existing jest one (for example)? I can see how some tasks might be independent (e.g. linting vs building), but compiling and testing feel closely connected to me because there's such a dependency between them. I'd be delighted to chat about it!

.github/workflows/typescript.yml
tjmw: Hey great question Adam! I don't think we do have any guidelines currently. For full disclosure I ~~stole~~ borrowed this from the DCR TypeScript action config as I was interested to see the impact it had (with the idea that if this was successful, we could migrate other actions over). I'd be up for switching over to the GitHub cache action as an experiment to see how that looks. Either before or after merging this.

tjmw:
Thanks for the reviews @michaelclapham & @adamnfish 👍 

> I'm interested to hear why is this in a separate workflow, instead of being part of the existing jest one (for example)?

In general I'd rather keep actions small/granular for similar reasons I like to keep individual tests as focused as possible - that it makes debugging in the case of a failure easier. I think it's possible for the TS pass and tests to fail for example (and vice versa) and I think having separate actions makes that easier to troubleshoot. What do you reckon?

.github/workflows/typescript.yml
tjmw: We chatted about this in the Client Side Infra meeting today. There aren't any established guidelines around actions and when to pick 3rd party or GitHub maintained. The action in question here is being used on other repos and there's a high level of trust around it. We talked about possibly maintaining a list of recommended actions for consistency across repos. One thing I learned is that we can [keep our actions up to date with Dependabot](https://docs.github.com/en/code-security/supply-chain-security/keeping-your-dependencies-updated-automatically/keeping-your-actions-up-to-date-with-dependabot) which is great. That brings the confidence we have in keeping our actions up to date in line with other library dependencies.

tjmw:
I'm going to go ahead and merge this as I think it's useful as is. I do think we should start using dependabot to keep our actions up to date. Let's see how we feel about keeping ts separate from jest etc. Adam suggested maybe using steps to share the setup between things like ts and jest (https://docs.github.com/en/actions/learn-github-actions/managing-complex-workflows).

PR #173, opened 2021-06-21
Integrate with GuTools login and image signing service

tjmw:
## What does this change?

* Makes the Grid integration visible by not blocklisting the `AppBanner` component.
* Create a signed image URL by calling the new [image signing service](https://github.com/guardian/image-url-signing-service) when an image URL is received from Grid.
* Checks that the user is logged in on the .gutools.co.uk domain and redirects to login.gutools.co.uk if not (you'll need to be logged in for requests to Grid and the image signing service to work).

## Accessibility

### DIFF
--- a/.storybook/grid/GridPanel.tsx
+++ b/.storybook/grid/GridPanel.tsx
@@ -3,10 +3,10 @@ import { AddonPanel } from '@storybook/components';
 import { useChannel } from '@storybook/api';
 import { STORY_CHANGED } from '@storybook/core-events';
 import { API } from '@storybook/api';
-import type { IframePostMessage } from '@guardian/grid-client';
-import { styles } from './gridStyles';
+import type { IframePostMessage, Asset } from '@guardian/grid-client';
+import { nonEmotionStyles } from './gridStyles';
 import { INITIAL_IMAGE_EVENT, IMAGE_SELECTED_EVENT } from './withGrid';
-import { GRID_URL } from '../utils';
+import { getGridUrl, getImageSigningUrl } from '../utils';
 
 const isValidMessage = (data: IframePostMessage) =>
     data?.crop?.data && data?.image?.data && data?.crop?.data?.assets?.length > 0;
@@ -16,6 +16,36 @@ type Props = {
     api: API;
 };
 
+// We think these params are the best combination across both high dpi and
+// standard dpi screens while keeping the image size down. In future we should
+// switch to using the picture element end generate different images depending
+// on the user's screen/device.
+const imageProfile = {
+    width: 930,
+    quality: 45,
+    auto: 'format',
+};
+
+const signImageUrl = (unsignedImageUrl: URL): Promise<string> =>
+    fetch(`${getImageSigningUrl()}/signed-image-url`, {
+        method: 'POST',
+        headers: {
+            'Content-Type': 'application/json',
+            Accept: 'application/json',
+        },
+        body: JSON.stringify({
+            url: unsignedImageUrl,
+            profile: imageProfile,
+        }),
+        credentials: 'include',
+    })
+        .then((res) => {
+            return res.json();
+        })
+        .then((json) => {
+            return json.signedUrl;
+        });
+
 export const GridPanel = ({ api, active }: Props) => {
     const emit = useChannel({});
     const [isModalOpen, setIsModalOpen] = useState<boolean>(false);
@@ -23,7 +53,7 @@ export const GridPanel = ({ api, active }: Props) => {
     const activeButton = useRef(null);
 
     const onMessage = (event) => {
-        if (event.origin !== GRID_URL) {
+        if (event.origin !== getGridUrl()) {
             return;
         }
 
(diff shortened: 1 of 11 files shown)


.storybook/grid/GridPanel.tsx
```diff
@@ -16,14 +16,31 @@ type Props = {
...
+type CropAsset = {
+    file: URL;
+    size?: number;
+    mimeType?: string | number;
+    dimensions?: { width: number; height: number; };
+    secureUrl?: URL;
+};
+
+function getBestCrop(cropAssets: CropAsset[]) {
+    const IDEAL_WIDTH = 500;
```
tjmw, line 28: This is definitely an improvement on the previous behaviour of just picking the last crop in the list. Not for this PR but it'd be interesting to think about how we parameterise this when we want to use the image picker for other components which might have different requirements (I think we might want to use this on the newsletter epic).

.storybook/grid/GridPanel.tsx
```diff
@@ -34,16 +51,36 @@ export const GridPanel = ({ api, active }: Props) => {
             return;
         }
 
-        const cropAssets = data.crop.data.assets;
-        const cropUrl = cropAssets[cropAssets.length - 1]?.secureUrl;
+        const cropAsset = getBestCrop(data.crop.data.assets.concat([]));
```
tjmw, line 54: What does the `concat` get us here?

.storybook/grid/GridPanel.tsx
```diff
@@ -34,16 +51,36 @@ export const GridPanel = ({ api, active }: Props) => {
...
+                "Content-Type": "application/json",
+                "Accept": "application/json"
+            },
+            body: JSON.stringify({
+                url: cropUrl,
+                profile: {
+                    width: 500
+                }
+            }),
+            credentials: "include"
```
tjmw, line 74: I wonder if it'd be worth extracting this out to a separate function?

.storybook/grid/gridStyles.ts
```diff
@@ -1,6 +1,99 @@
-import { css } from '@emotion/core';
+import { css } from '@emotion/react';
```
tjmw, line 1: Presumably this import isn't used any more? I wonder why the linter isn't shouting about that 🤔

.storybook/grid/gridStyles.ts
```diff
@@ -1,6 +1,99 @@
...
+        borderRadius: '4px',
+        lineHeight: '20px',
+        flex: '1',
+        textAlign: 'left',
+        overflow: 'visible',
+        height: '52px',
+    },
+};
 
 export const styles = {
```
tjmw, line 98: Oh I guess these emotion styles can go now we're not using emotion in this context?

.storybook/gu-auth/GuAuth.tsx
```diff
@@ -0,0 +1,101 @@
...
+
+    useEffect(() => {
+        const handleLogin = async () => {
+            const redirectToLogin = () => {
+                const currentLocation = window.location.href;
+
+                const newHref = `${getLoginUrl()}/login?returnUrl=${encodeURIComponent(
+                    currentLocation,
+                )}`;
+                console.log('Redirecting to ', newHref);
```
tjmw, line 65: Let's remove this `console.log`.

.storybook/grid/GridPanel.tsx
michaelclapham: I think before the array is a weird read only type which for some reason doesn't have the sort method on it.

.storybook/grid/gridStyles.ts
tjmw: I went ahead and removed these.

.storybook/grid/GridPanel.tsx
tjmw: Ah cool thanks for explaining 👍  I wonder if we should add a comment saying the same thing.

.storybook/grid/gridStyles.ts
```diff
@@ -1,108 +1,95 @@
...
-        border-radius: 4px;
-        border-bottom-style: none;
-        padding: 10px 16px;
-        display: inline;
-        overflow: visible;
-        &:focus {
-            outline: none !important;
-        }
-    `,
+export const nonEmoStyles: Record<string, React.CSSProperties> = {
```
arausuy, line 5: I'm guessing this is non-emotion styles?

arausuy: The variable name makes me want to rebel and just buy clothes from hot topic

.storybook/gu-auth/GuAuth.tsx
```diff
@@ -0,0 +1,101 @@
...
+            }
+        }
+    }, [hasCheckedLoggedInStatus]);
+
+    useEffect(() => {
+        const handleLogin = async () => {
+            const redirectToLogin = () => {
+                const currentLocation = window.location.href;
+
+                const newHref = `${getLoginUrl()}/login?returnUrl=${encodeURIComponent(
```
arausuy, line 62: The naming could be improve slightly here to sign post that it is a redirecting to a login

.storybook/grid/GridPanel.tsx
arausuy: Would a comment be useful to explain how we've settled on 500?

tjmw: Yeah that's a great shout, partly because it would force me to remember the reasoning!

.storybook/grid/GridPanel.tsx
tjmw: I've added a comment explaining this. I also updated `getBestCrop` to use the `Asset` type from `@guardian/grid-client` instead of (re)defining it ourselves.

.storybook/grid/gridStyles.ts
tjmw: Yeah correct, something in the build pipeline was messing up handling the `css` prop. It'd be nice to fix that at some point, but not using emotion to style the storybook panel is fine and gets us moving.

.storybook/gu-auth/GuAuth.tsx
tjmw: 👍  I updated the name of the const.

.storybook/grid/GridPanel.tsx
tjmw: I went ahead and did this.

.storybook/grid/gridStyles.ts
michaelclapham: > The variable name makes me want to rebel and just buy clothes from hot topic

This made me think of the fantastic Emo the musical, and now I'm slightly sad: https://www.theguardian.com/film/2016/aug/15/emo-the-musical-review-tale-of-tolerance-doomed-by-flat-songs-and-infantile-jokes

michaelclapham: Renamed to nonEmotionStyles

.storybook/grid/GridPanel.tsx
tjmw: This was a really great question to ask! Having spoken to folks more familiar with the Fastly image service than I am, it seems like we should always be using the `master` crop as the source (which makes sense, since Fastly is able to do the resizing, so we don't need to rely on the Grid to create different crop widths for us).

PR #175, opened 2021-06-25
US Newsletter Epic component

michaelclapham:
Start of US Newsletter Epic.

Trello Card: https://trello.com/c/csMyMtNN/1162-static-version-of-us-first-thing-newsletter-epic﻿

### DIFF
--- a/src/BrazeMessageComponent.tsx
+++ b/src/BrazeMessageComponent.tsx
@@ -15,6 +15,8 @@ import {
 } from './TheGuardianIn2020Banner';
 import { BrazeClickHandler } from './utils/tracking';
 
+import { COMPONENT_NAME as US_NEWSLETTER_EPIC_NAME, USNewsletterEpic } from './USNewsletterEpic';
+
 type BrazeMessageProps = {
     [key: string]: string | undefined;
 };
@@ -34,6 +36,7 @@ const COMPONENT_MAPPINGS: ComponentMapping = {
     [APP_BANNER_NAME]: AppBanner,
     [SPECIAL_EDITION_BANNER_NAME]: SpecialEditionBanner,
     [THE_GUARDIAN_IN_2020_BANNER_NAME]: TheGuardianIn2020Banner,
+    [US_NEWSLETTER_EPIC_NAME]: USNewsletterEpic,
 };
 
 export type Props = {
--- a/src/USNewsletterEpic/index.tsx
+++ b/src/USNewsletterEpic/index.tsx
@@ -67,6 +67,9 @@ const styles = {
     button: css`
         background-color: #c70000;
         color: #ffffff;
+        &:hover {
+            background-color: #c70000;
+        }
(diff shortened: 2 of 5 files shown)


src/BrazeMessageComponent.tsx
```diff
@@ -34,6 +36,7 @@ const COMPONENT_MAPPINGS: ComponentMapping = {
     [APP_BANNER_NAME]: AppBanner,
     [SPECIAL_EDITION_BANNER_NAME]: SpecialEditionBanner,
     [THE_GUARDIAN_IN_2020_BANNER_NAME]: TheGuardianIn2020Banner,
+    [US_NEWSLETTER_EPIC_NAME]: USNewsletterEpic,
```
tjmw, line 39: I think we've agreed to pull this out into a separate factory component specific to `EndOfArticle` which would have it's own mappings. Something like `BrazeEndOfArticleComponent`. I think it's fine to do that in a separate PR though.

src/USNewsletterEpic/index.tsx
```diff
@@ -0,0 +1,134 @@
...
+                fillRule="evenodd"
+                clipRule="evenodd"
+                d="M26 15c0 6.075-4.925 11-11 11S4 21.075 4 15 8.925 4 15 4s11 4.925 11 11zm-9.8-.35L15.475 6H14.5l-.75 9.375 1.275 1.275L22 16v-1l-5.8-.35z"
+            />
+        </svg>
+    );
+};
+
+const IMAGE_URL =
+    'https://i.guim.co.uk/img/media/d0944e021b1cc7426f515fecc8034f12b7862041/0_0_784_784/784.png?width=196&s=fbdead3f454e1ceeeab260ffde71100a';
```
tjmw, line 30: A thing I learned from talking to folks who work on the Grid recently is that we should always be using the `master` crop as the source image and letting the image service do any resizing. Let's update this accordingly.

src/USNewsletterEpic/index.tsx
```diff
@@ -0,0 +1,134 @@
...
+                    <div css={styles.frequencySection}>
+                        <span css={styles.frequencyClock}>
+                            <SvgClock />
+                        </span>
+                        <span css={styles.frequencyText}>{frequency}</span>
+                    </div>
+                    <p css={commonStyles.paragraph}>{paragraph1}</p>
+                    {paragraph2 ? <p css={commonStyles.paragraph}>{paragraph2}</p> : null}
+                    <ThemeProvider theme={buttonBrandAlt}>
+                        <Button css={styles.button}>Sign up</Button>
```
tjmw, line 126: Would be interesting to know what we'd like to do with this button when it comes to tracking (I think we can put this off until we add the functionality to actually do the signup). I guess we'd want to log the click to Ophan/Braze.

tjmw (approved):
This looks great, nice work!

PR #183, opened 2021-07-05
Use head request to check local login service

michaelclapham:
Use HEAD request when running locally to check to see if local login service is online.
If not online then don't require login in storybook.﻿


tjmw:
Closing for now, I think the current approach is working okay.

PR #186, opened 2021-07-09
UK and Australia Today Newsletter Epic components

michaelclapham:
## What does this change?
This converts the USNewsletterEpic component into a generic NewsletterEpic component, and adds 3 components that use the newsletter epic.
- UK Morning Briefing
- US First Thing
- Australia Morning Briefing

This components is still a static component. That is, clicking Sign up will not sign up the user to the newsletter.

## How to test
Visit braze-components.code.gutools.co.uk

## Images
Screenshot 2021-07-13 at 18 02 44

## Accessibility

- [x] [Tested with screen reader](https://www.theguardian.design/2a1e5182b/p/6691bb-accessibility/t/558398)
- [x] [Navigable with keyboard](https://www.theguardian.design/2a1e5182b/p/6691bb-accessibility/t/39894d)
- [x] [The change doesn't use only colour to convey meaning](https://www.theguardian.design/2a1e5182b/p/6691bb-accessibility/t/29032f)


src/AUNewsletterEpic/index.stories.tsx
```diff
@@ -0,0 +1,50 @@
+import React, { ReactElement } from 'react';
+import { BrazeMessageComponent } from '../BrazeMessageComponent';
+import { StorybookWrapper } from '../utils/StorybookWrapper';
+import { withKnobs, text } from '@storybook/addon-knobs';
+
+export default {
+    component: 'AUNewsletterEpic',
+    title: 'WorkInProgress/EndOfArticle/AUNewsletterEpic',
+    decorators: [withKnobs({ escapeHTML: false })],
```
tjmw, line 9: I'm not sure we need the `escapeHTML: false` here. I think this is needed when you expect to pass HTML through to the component and render it as-is, which I don't think we do here.

src/AUNewsletterEpic/index.tsx
```diff
@@ -0,0 +1,30 @@
+import React from 'react';
+import { canRender, COMPONENT_NAME } from './canRender';
+import { NewsletterEpic, BrazeMessageProps } from '../NewsletterEpic';
+import { BrazeClickHandler } from '../utils/tracking';
+import { OphanComponentEvent } from '@guardian/types';
+
+const IMAGE_URL =
+    'https://i.guim.co.uk/img/media/9f9f9c06ed5a323b13be816d5c160728c81d1bf9/0_0_784_784/784.png?width=196&s=4c4d5ff2c20821a6e6ff4d25f8ebcc16';
```
tjmw, line 8: I think we should be using the `master` crop as the source and then letting the image resizer do its job. (So I'd expect the source URL to have `master` in it somewhere - I think we're using a Grid crop rather than the master crop here).

src/BrazeMessageComponent.tsx
```diff
@@ -15,8 +15,14 @@ import {
 } from './TheGuardianIn2020Banner';
 import { BrazeClickHandler } from './utils/tracking';
 
+import { COMPONENT_NAME as NEWSLETTER_EPIC_NAME, NewsletterEpic } from './NewsletterEpic';
```
tjmw, line 18: I wonder - does it make sense to have the `NewsLetterEpic` be exposed as a component that's usable by marketing at this point (same question about the Storybook story for this one). It feels like the final piece to having a truly generic component is to be able to use the image picker with it. What do you reckon?

src/UKNewsletterEpic/index.stories.tsx
```diff
@@ -0,0 +1,49 @@
+import React, { ReactElement } from 'react';
+import { BrazeMessageComponent } from '../BrazeMessageComponent';
+import { StorybookWrapper } from '../utils/StorybookWrapper';
+import { withKnobs, text } from '@storybook/addon-knobs';
+
+export default {
+    component: 'UKNewsletterEpic',
+    title: 'WorkInProgress/EndOfArticle/UKNewsletterEpic',
+    decorators: [withKnobs({ escapeHTML: false })],
```
tjmw, line 9: As above, not sure we need the `escapeHTML` here.

src/USNewsletterEpic/index.tsx
```diff
@@ -1,92 +1,12 @@
...
-    `,
-};
-
-export type BrazeMessageProps = {
-    header?: string;
-    frequency?: string;
-    paragraph1?: string;
-    paragraph2?: string;
-};
-
```
tjmw, line 89: Nice to be able to move this all up to the parent component! 🎉

tjmw (commented):
Nice work, this is looking great 👍  I left a few comments/questions, see what you think. Happy to chat about any of the comments if it's useful.

src/BrazeMessageComponent.tsx
michaelclapham: I've removed it from Storybook for this PR. In addition to the grid integration, we would probably want a user friendly way to select the newsletter id, so I think making it generic for marketing deserves it's own PR.

src/UKNewsletterEpic/index.stories.tsx
michaelclapham: Without it we can't support the apostrophe in the title. It get's replaced with &apos;

Is there a better way to "unescape" this?

tjmw: Discussed offline - it sounds like we do need this. My bad, sorry!

PR #188, opened 2021-07-13
Fix broken epic rendering

tjmw:
## What does this change?

The ContributionsEpic component added some validation to the passed props which we weren’t passing here. Add the missing props in so the component renders correctly.

## How to test

`yarn storybook`

The epic component renders as expected in storybook.

## Images

Before (no epic):

Screenshot 2021-07-13 at 15 28 32

After (an epic):

Screenshot 2021-07-13 at 15 28 22

## Accessibility

### DIFF
--- a/src/Epic/index.stories.tsx
+++ b/src/Epic/index.stories.tsx
@@ -47,9 +47,14 @@ declare global {
 type EpicProps = {
     variant: Variant;
     tracking: Record<string, unknown>;
+    articleCounts: {
+        for52Weeks: number;
+        forTargetedWeeks: number;
+    };
 };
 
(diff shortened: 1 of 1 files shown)


michaelclapham (approved):

src/Epic/index.stories.tsx
```diff
@@ -176,6 +195,7 @@ export const defaultStory = (): ReactElement | null => {
     };
 
     const epicProps = buildProps(knobs);
+    console.log({ epicProps });
```
tjmw, line 198: Oops, I didn't mean for this to slip in. Will remove.

PR #191, opened 2021-07-16
Added our own version of Contributions Epic

michaelclapham:
## What does this change?
This adds new Epic component that is a visual clone of the ContributionsEpic from support-dotcom-components.

It does not have any of the ophan tracking, or article count support, but it does support currency and country code substitutions. 

## How to test
Compare ExternalEpic (the old component) in storybook, with Epic (the new component).

Epic component looks the same and users who see an epic delivered via a Braze message don't notice any difference.

*The UI test for Epic passed with no differences when switching from the external to internal component*

Loss of tracking data. Lack of article count not communicated to marketing.


index.ts
```diff
@@ -1 +1,5 @@
 export * from './src/BrazeMessageComponent';
+
+// Temporary to fix css prop error in Epic story
```
tjmw, line 3: Oh, I'm surprised this is still needed. Isn't the new epic component now imported by `BrazeMessageComponent`?

src/BrazeMessageComponent.tsx
```diff
@@ -46,6 +48,7 @@ const COMPONENT_MAPPINGS: ComponentMapping = {
     [US_NEWSLETTER_EPIC_NAME]: USNewsletterEpic,
     [AU_NEWSLETTER_EPIC_NAME]: AUNewsletterEpic,
     [UK_NEWSLETTER_EPIC_NAME]: UKNewsletterEpic,
+    Epic: Epic,
```
tjmw, line 51: Shall we import the name here from the component file, in the same way we do for the other components?

src/Epic/canRender.ts
```diff
@@ -6,10 +6,11 @@ import { BrazeMessageProps } from './index';
 export const canRenderEpic = (brazeMessageProps: BrazeMessageProps): boolean => {
     const { buttonText, buttonUrl, ophanComponentId } = brazeMessageProps;
     const paragraphs = parseParagraphs(brazeMessageProps);
+    console.log(buttonText, buttonUrl, ophanComponentId, paragraphs);
```
tjmw, line 9: Let's remove this `console.log`.

src/Epic/canRender.ts
```diff
@@ -6,10 +6,11 @@ import { BrazeMessageProps } from './index';
...
     return Boolean(buttonText && buttonUrl && ophanComponentId && paragraphs.length > 0);
 };
 
-const parseParagraphs = (brazeMessageProps: BrazeMessageProps): string[] => {
+export const parseParagraphs = (brazeMessageProps: BrazeMessageProps): string[] => {
```
tjmw, line 13: Now that we're not tied to the interface of the RRCP `ContributionsEpic`, I wonder if it makes sense to have the interface of our component more closely match the way the data will be in Braze (i.e. `paragraph1`, `paragraph2` etc). What do you reckon?

src/Epic/contrib.index.tsx
```diff
@@ -0,0 +1,217 @@
...
+    countryCode,
+    submitComponentEvent,
+}: EpicProps) => {
+    const [hasBeenSeen, setNode] = useHasBeenSeen({ threshold: 0 }, true) as HasBeenSeen;
+
+    useEffect(() => {
+        if (hasBeenSeen) {
+            // sendEpicViewEvent(tracking.referrerUrl, stage);
+        }
+    }, [hasBeenSeen]);
```
tjmw, line 180: I think we can probably remove this `useEffect`, and also supporting stuff like `hasBeenSeen` from here (I think this is all related to reporting epic views back to a system that the contributions team use).

src/Epic/contrib.index.tsx
```diff
@@ -0,0 +1,217 @@
...
+        }
+    }
+
+    if (
+        [cleanHighlighted, cleanHeading, ...cleanParagraphs].some(
+            containsNonArticleCountPlaceholder,
+        )
+    ) {
+        console.warn('braze-components Epic does not support article count');
+        return null; // quick exit if something goes wrong. Ideally we'd throw and caller would catch/log but TODO that separately
```
tjmw, line 200: Hmm if these checks result in us not rendering the component, then they should probably be part of the epic `canRender`. What do you think?

src/Epic/contrib.index.tsx
```diff
@@ -0,0 +1,217 @@
...
+            cleanParagraphs.push(paragraph);
+        }
+    }
+
+    if (
+        [cleanHighlighted, cleanHeading, ...cleanParagraphs].some(
+            containsNonArticleCountPlaceholder,
+        )
+    ) {
+        console.warn('braze-components Epic does not support article count');
```
tjmw, line 199: I _think_ when the condition of this `if` block is met it means there are placeholders (like `%%SOMETHING%%`) in the text which _aren't_ related to `%%ARTICLE_COUNT%%`. I'm not sure why it's okay for `%%ARTICLE_COUNT%%` to still exist in the text, it seems to be handled separately than other placeholders.

src/Epic/index.stories.tsx
```diff
@@ -75,63 +45,25 @@ type DataFromKnobs = {
...
         },
+        ophanComponentId: data.ophanComponentId,
+        countryCode: 'GB',
+        submitComponentEvent: () => console.log('Submit component event'),
+        logButtonClickWithBraze: () => console.log('Log button click with Braze'),
     };
+    data.paragraphs.forEach((paragraph, index) => {
+        props.brazeMessageProps[`paragraph${index + 1}`] = paragraph;
+    });
+    console.log(data.paragraphs, props.brazeMessageProps);
```
tjmw, line 65: Let's remove the `console.log` from here.

src/Epic/index.stories.tsx
```diff
@@ -207,11 +129,9 @@ export const defaultStory = (): ReactElement | null => {
         return null;
     }
 
-    const Component = getComponentfromName(componentName);
-
     return (
         <StorybookWrapper>
-            <Component {...epicProps} />
+            <Epic {...epicProps}></Epic>
```
tjmw, line 134: I think we should be going through the component factory `BrazeMessageComponent` here to be consistent with the other stories (it's also closer to how we'll consume this in DCR).

src/Epic/index.tsx
```diff
@@ -0,0 +1,118 @@
...
+                    <span css={styles.heading}>{heading}</span>
+                    {paragraphs.map((text, index) => (
+                        <p key={'paragraph' + index} css={styles.paragraph}>
+                            {text}
+                            {index === paragraphs.length - 1 ? (
+                                <span css={styles.highlightText}>{highlightTextClean}</span>
+                            ) : null}
+                        </p>
+                    ))}
+                    {/* buttonText and buttonUrl will have been checked for not undefined in canRenderEpic */}
```
tjmw, line 109: 👍

src/ExternalEpic/canRender.ts
```diff
@@ -0,0 +1,26 @@
+import { BrazeMessageProps } from './index';
```
tjmw, line 1: I guess we'll be able to get rid of `ExternalEpic` once we've moved over to using the new component from this lib in DCR?

src/Epic/contrib.index.tsx
```diff
@@ -0,0 +1,217 @@
+import { css } from '@emotion/react';
```
tjmw, line 1: Oh I left a couple of comments in this file, but I realised that it's maybe not used anywhere, is that right? If so please ignore comments I left in here!

src/Epic/index.tsx
```diff
@@ -0,0 +1,118 @@
...
+
+    if (!canRenderEpic(props.brazeMessageProps)) {
+        return null;
+    }
+
+    const paragraphs = parseParagraphs(props.brazeMessageProps).map((paragraph) =>
+        replaceNonArticleCountPlaceholders(paragraph, countryCode),
+    );
+
+    const highlightTextClean = replaceNonArticleCountPlaceholders(highlightedText, countryCode);
```
tjmw, line 94: There is a code path in `replaceNonArticleCountPlaceholders` where we fail to replace `%%COUNTRY_NAME%%` because we can't derive the country name from the country code passed. Which means there _could_ be unreplaced placeholders here. I wonder how important that is.

It'd be interesting to know which (if any) or these placeholders might be useful (or already used) by marketing. Pretty sure country name isn't currently.

We could return `null` early here, like the contributions epic does. But if we do that I think we'd want to add the check to our `canRender`, but that increases the complexity of `canRender` quite a bit as we'd need to have access to the country code, and try to do the replacement in there, which seems overly complex.

What do you reckon? Maybe it would be better to do any replacement like this using liquid template logic in Braze, that way we wouldn't have to support placeholders at all (we should check first if we're already using them for existing braze epic campaigns).

tjmw (commented):
Hey this is looking good 🥇  I left quite a few comments but most are pretty minor. Have a read and see what you think. Happy to discuss next week. I think the most interesting/complex bit is around the placeholder logic.

src/Epic/canRender.ts
michaelclapham: Done

src/Epic/canRender.ts
michaelclapham: I agree, however this function is still useful as it allows us to go from paragraph1, paragraph2 to a list of paragraphs, which makes the render and canRender functions simpler.

src/Epic/index.stories.tsx
michaelclapham: Done

src/Epic/index.tsx
michaelclapham: I think I agree that using Liquid logic is the right way to go here. It does seem silly to build our own version considering there's nothing that we support that can't be done via Liquid.

src/ExternalEpic/canRender.ts
michaelclapham: Yes I think so. Although I was wondering if there is anyway to setup the Chromatic tests to compare the two components?

src/BrazeMessageComponent.tsx
michaelclapham: Done

src/ExternalEpic/canRender.ts
tjmw: Ooh that's an interesting thought. I'm not sure - Chromatic generally compares an accepted baseline version of a component with a later iteration of the same component. Would be cool to spend some minutes on figuring out what's possible. This could be a really cool way of avoiding drift between the two components.

src/Epic/index.tsx
tjmw: Yeah I guess the dealbreaker will be the data we want to use to use to drive the templating. For example the country Braze knows about (which is what we'd have to use in the Liquid) is potentially different than the country in the browser which comes from Fastly's geolocation service.

index.ts
michaelclapham: This is for the old external epic. Previously it wasn't imported anywhere so had the css prop error. The css prop error only shows up in VSCode and doesn't affect the Github / TeamCity build so we could remove this.

src/Epic/placeholders.ts
```diff
@@ -0,0 +1,30 @@
...
+    countryCode?: string,
+): string => {
+    if (!content) {
+        return '';
+    }
+
+    content = content.replace(/%%CURRENCY_SYMBOL%%/g, getLocalCurrencySymbol(countryCode));
+
+    const countryName = getCountryName(countryCode) ?? '';
+    content = countryName ? content.replace(/%%COUNTRY_NAME%%/g, countryName) : content;
```
tjmw, line 18: There is a scenario here where `%%COUNTRY_NAME%%` won't get replaced (if `countryName` isn't truthy). I don't think we'd want to render the epic if it hasn't been replaced. I wonder if we should remove `%%COUNTRY_NAME%%` from the list of supported placeholders since I don't _think_ it's used currently and it would allow us to sidestep the issue. The alternative would be to come up with a default countryName` and I'm not sure there's one which makes sense. What do you think?

index.ts
tjmw: Ah I see, thanks for explaining 👍  Strangely I tried removing this locally and don't see the error (it was in the `ExternalEpic` story right and `yarn tsc` also seems happy. I guess I'm slightly nervous here about `import`ing a story file in one of our main entry points, does this end up adding the story to the prod/npm bundle? Separately I definitely think we should remove the `console.log` just below.

src/Epic/placeholders.ts
```diff
@@ -0,0 +1,30 @@
...
+
+    return content;
+};
+
+// Nb. don't attempt to use lookbehind (?<!) here, as IE 11 will break alas
+const PLACEHOLDER_RE = /%%.*?%%/g;
+export const containsNonAllowedPlaceholder = (text: string): boolean => {
```
tjmw, line 25: Nice! What do you think about adding a test for this function? I think it could be helpful and would act as documentation.

src/Epic/index.stories.tsx
```diff
@@ -1,28 +1,19 @@
...
-import * as emotionReact from '@emotion/react';
-import * as emotionReactJsxRuntime from '@emotion/react/jsx-runtime';
-
-const css = emotionReact.css;
+import { BrazeMessageComponent } from '../BrazeMessageComponent';
 
 export default {
     component: 'Epic',
-    title: 'EndOfArticle/Epic',
+    title: 'WorkInProgress/EndOfArticle/Epic',
```
tjmw, line 8: 👍  Great idea to add the new epic to `WorkInProgress`

tjmw (commented):
Nice! This is looking really good. I left a few comments, nothing major. The only one I'd love to figure out before merging is that import of the story in the main `index.ts`, really happy to pair on that to either figure out if there's an alternative or prove that it isn't having an adverse affect on the bundle size.

PR #197, opened 2021-07-23
Remove margins from both epics

tjmw:
## What does this change?

Removes some horizontal padding from both newsletter and contributions epic components. We noticed when rendered in the context of an article there was some extra whitespace we didn't want.

## Images

Newsletter epic example (note the spacing to the left and how the left edge of the component aligns with the article text):

### Before

Screenshot 2021-07-23 at 09 29 19

### After

Screenshot 2021-07-23 at 09 27 03

### DIFF
--- a/src/Epic/index.tsx
+++ b/src/Epic/index.tsx
@@ -23,7 +23,6 @@ const linkStyles = css`
 const styles = {
     epicWrapper: css`
         max-width: 620px;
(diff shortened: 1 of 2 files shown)


arausuy (approved):

tjmw:
Thanks @arausuy! 👍

PR #203, opened 2021-07-27
New epic bundle with same contents as main bundle

michaelclapham:
## What does this change?
Splits banners and end of article components (epics) into separate bundles, to reduce file size lazy loaded by users.

## How to test
Run yarn build and have a look at the bundle sized produced. More testing in DCR possible once new version published to npm.

Main bundle size not increased by 20% (which it was when it included new epics). https://github.com/guardian/dotcom-rendering/pull/3209#issuecomment-886886242

Loading more code when user loads both banners and epics.


arausuy:
I really like the explicitness of having a set of banners and a set of under article options!

arausuy (approved):

PR #208, opened 2021-07-30
Fix invalid placeholder detection logic

tjmw:
## What does this change?

Fix the invalid placeholder detection logic and also added some tests as documentation. Previously the code to determine
whether there were any invalid placeholders was always returning false, this fixes the check so that invalid placeholders are correctly flagged.

The fields we support placeholder replacement in are the paragraphs and highlightedText. Update the call to `containsNonAllowedPlaceholder` in the canRender to reflect this.

## How to test

`yarn test`

You can also try out different placeholders in storybook. If an invalid one is used in the paragraphs or highlighted text the component won't render.


michaelclapham (approved):
👍

PR #215, opened 2021-08-16
Loading animation for newsletter epic

michaelclapham:
## What does this change?
Adds animation for request in progress for when Sign up is clicked on the newsletter epic.

## How to test
Open storybook and select a newsletter epic, then click Sign up. For one second the loading animation will show.

## Accessibility

### DIFF
--- a/src/NewsletterEpic/LoadingDots.tsx
+++ b/src/NewsletterEpic/LoadingDots.tsx
@@ -1,0 +1,80 @@
+import { css, keyframes } from '@emotion/react';
+import React from 'react';
+
+const loadingAnimKeyframes = keyframes`
+    0% {
+        transform: scale(1);
+        filter: brightness(1);
+    }
+    15% {
+        transform: scale(1.333);
+        filter: brightness(0.7);
+    }
+    30% {
+        transform: scale(1);
+        filter: brightness(1);
+    }
+`;
+
+const loadingAnimCss = css`
+    circle {
+        animation: ${loadingAnimKeyframes} 1.5s ease infinite;
+    }
+
+    #dot_1 {
+        animation-delay: 0ms;
+        transform-origin: 3px 3.5px;
+    }
+
+    #dot_2 {
+        animation-delay: 400ms;
+        transform-origin: 17.4px 3.5px;
+    }
+
+    #dot_3 {
+        animation-delay: 800ms;
+        transform-origin: 31.7px 3.5px;
+    }
+`;
+
+export const LoadingDots = (): React.ReactElement => {
+    return (
+        <svg
+            width="50"
+            height="17"
+            viewBox="-5 -5 45 12"
+            fill="none"
+            xmlns="http://www.w3.org/2000/svg"
+            css={loadingAnimCss}
+        >
+            <g id="Dots step 1">
+                <g id="Group 660">
+                    <circle
+                        id="dot_1"
+                        opacity="0.5"
+                        cx="3.0152"
+                        cy="3.56641"
+                        r="3"
+                        fill="#707070"
+                    />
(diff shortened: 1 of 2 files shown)


src/NewsletterEpic/LoadingDots.tsx
```diff
@@ -0,0 +1,80 @@
+import { css, keyframes } from '@emotion/react';
+import React from 'react';
+
+const loadingAnimKeyframes = keyframes`
```
tjmw, line 4: Nice work getting this implemented with css!

src/NewsletterEpic/index.tsx
```diff
@@ -169,21 +174,24 @@ const CTA: React.FC<CTAProps> = (props: CTAProps) => {
...
-        case 'IN_PROGRESS':
+        case 'FAILURE':
             return (
-                <ThemeProvider theme={buttonBrandAlt}>
-                    <Button css={ctaStyles.button} disabled={true}>
-                        Loading...
-                    </Button>
-                </ThemeProvider>
+                <div css={ctaStyles.errorText}>
+                    There was an error signing up to the newsletter. Please try agian
```
tjmw, line 190: Tiny typo here:

```suggestion
                    There was an error signing up to the newsletter. Please try again
```

tjmw, line 190: Here we're asking the user to try again, but it doesn't look like we're giving them an easy way to do that, should we re-render the button? Or make `Try again` clickable?

tjmw (commented):
This is looking good, I just left a couple of comments/questions.

src/NewsletterEpic/index.tsx
michaelclapham: I've gone for re-render the sign up button and moved that button into it's own component. Do you reckon we're okay with the amount this changes the height of the epic though?

tjmw: Good question. All else being equal I guess it'd be nice if it didn't increase the height too much but since we expect this error case to be fairly exceptional I don't think we should worry too much.

src/NewsletterEpic/index.tsx
```diff
@@ -151,39 +157,49 @@ const ctaStyles = {
...
+        subscribeToNewsletter(newsletterId as string)
+            .then(() => setSubscribeClickStatus('SUCCESS'))
+            .catch(() => {
+                setSubscribeClickStatus('FAILURE');
+            });
+    };
+
     switch (subscribeClickStatus) {
         case 'DEFAULT':
+            return <SignUpButton onSignUpClick={onSignUpClick}></SignUpButton>;
```
tjmw, line 191: Purely subjective but what do you think about writing this using a self-closing tag? I tend to do that when the component has no children.

```suggestion
            return <SignUpButton onSignUpClick={onSignUpClick} />;
```

src/UKNewsletterEpic/index.stories.tsx
```diff
@@ -39,7 +39,9 @@ export const defaultStory = (): ReactElement | null => {
                 }}
                 subscribeToNewsletter={(newsletterId) => {
                     console.log(`subscribeToNewsletter invoked with id ${newsletterId}`);
-                    return new Promise((resolve) => setTimeout(resolve, 1000));
+                    return new Promise((resolve, reject) =>
+                        setTimeout(() => (Math.random() > 0.5 ? resolve() : reject()), 1000),
```
tjmw, line 43: I reckon rather than have this succeed or fail randomly we should have separate stories for success and failure, what do you think?

tjmw (approved):
Looking good! I just left a couple of minor comments/suggestions.

PR #218, opened 2021-08-24
Fix Storybook Gu Preview link for Epic

tjmw:
## What does this change?

For components in storybook we offer the ability to preview in place on the preview site via a custom Storybook plugin. This requires a little bit of wiring and it turns out we were missing it for the `Epic` component (I noticed this while attempting to troubleshoot an issue, which this would have made easier).

## How to test

Run storybook and click on the preview icon:


michaelclapham (approved):
Looks good!

PR #224, opened 2021-09-06
Migrate from `@guardian/types` to `@guardian/libs`

jamie-lynch:
## What does this change?

This PR migrates from the [`@guardian/types`](https://github.com/guardian/types) library to [`@guardian/libs`](https://github.com/guardian/libs) as [`@guardian/types` is now deprecated](https://github.com/guardian/types/pull/140).

## How to test

Run all validation and ensure that everything still runs as expected.


tjmw (commented):
Hi @jamie-lynch thanks for opening the PR. I think this looks good! We have `@guardian/libs` as a `devDependency` and `peerDependency`. At the moment the dev version string has been bumped as part of this PR but not the peer version, so I think we should update that too so that the two are in sync. I'm happy to pick up that change if it's useful? The tests/typescript compiler are happy so I think we're good, but I'll do a quick manual test too when I get a moment.

jamie-lynch:
> At the moment the dev version string has been bumped as part of this PR but not the peer version, so I think we should update that too so that the two are in sync.

Ah yes, good spot! I've updated the peerDependency version to match

tjmw:
Thanks @jamie-lynch 👍  Do you have a timeline for this work? I noticed the DCR PR linked to above. I'm just wondering when a good time to merge this/create a release would be. Since we're moving to specifying libs v3 as a peer dep I feel like we want to co-ordinate with DCR going to v3 (and roll the braze-components bump out with that in guardian/dotcom-rendering#3375, which I think is already your plan?) This lib is also used by frontend, so I think the same thing probably needs to happen there too. I guess the situation I want to avoid is creating a new release for this change and to then be blocked from being able to push out other changes to prod until the DCR libs v3 switch is ready. Any thoughts?

jamie-lynch:
Agreed that we need to coordinate this so that it doesn't become a blocker. The other libraries that are used by DCR and AR, and so need a new release before DCR/AR can be updated, are `atoms-rendering`, `image-rendering` and `discussion-rendering`. Once we get those PRs ready to merge, and we're confident that the only changes left to make on DCR and AR are bumping the deps then we should be able to release new versions of all the libraries at the same time and finish off the DCR/AR PRs.

With Frontend, I've opened this PR which seems to be all good without a new release of `braze-components`. Would you expect that to break until we've shipped this update?

tjmw:
Cool, yeah that sounds like a good plan 👍 

> With Frontend, I've opened this PR which seems to be all good without a new release of braze-components. Would you expect that to break until we've shipped this update?

Good question. I think probably I'd expect it to be okay but I'd like to test to make sure. The code around `@guardian/braze-components` in frontend hasn't been converted to TypeScript (yet) so I wouldn't expect any complaints from the compiler there (so that one change where we now get `OphanComponentEvent` from libs instead of types isn't relevant for frontend). I'd expect yarn to complain about mismatching peer dependency versions of libs (as the platform is providing v3 but the current release of braze-components expects v2) but for it to work as the surface area of libs we use is very small, just `storage`, which I don't think has changed at all from v2 to v3.

PR #222, opened 2021-08-27
Add EpicWithHeaderImage component

tjmw:
## What does this change?

This PR adds a new component `EpicWithHeaderImage`. The new component is implemented on top of the existing `Epic` component, passing through a hardcoded image URL. The image URL isn't part of the Braze message props in the existing `Epic`, so there's no way to pass it through from Braze (i.e. it's private for now, which is what we want).

We’re treating our first usage of the epic with an image as an experiment, so short term we’ll hardcode the image avoiding the overhead of adding proper image support (via the picker). If the test goes well we can make headerImageUrl a properly supported part of the `Epic` component.

https://trello.com/c/xODg9J4r/1217-new-epic-template-with-image

## How to test

In Storybook, there's a new story `EpicWithHeaderImage` under `WorkInProgress/EndOfArticle`.

## How can we measure success?

We'll be running a test to validate the image usage.

## Images

Actual image TBC (sadly the cats aren't it)!

Screenshot 2021-09-08 at 10 58 40

Screenshot 2021-08-27 at 16 44 06

## Accessibility

- [ ] [Tested with screen reader](https://www.theguardian.design/2a1e5182b/p/6691bb-accessibility/t/558398)
- [ ] [Navigable with keyboard](https://www.theguardian.design/2a1e5182b/p/6691bb-accessibility/t/39894d)
- [ ] [Colour contrast passed](https://www.theguardian.design/2a1e5182b/p/6691bb-accessibility/t/92b913)
- [ ] [The change doesn't use only colour to convey meaning](https://www.theguardian.design/2a1e5182b/p/6691bb-accessibility/t/29032f)

PR #231, opened 2021-09-14
Add EpicWithSpecialHeader

tjmw:
## What does this change?

This PR adds a new component `EpicWithHeaderImage`. The new component is implemented on top of the existing Epic component and adds a new header section including a headshot and some additional text.

This PR supersedes #222. It turned out the image marketing wanted to use was very text heavy, which didn't feel like a great fit (accessibility concerns, as well as general inflexibility of delivering text in an image).

Note: The entire new top section is hidden on mobile.

We’re treating our first usage of the epic with an image as an experiment, so short term we’ll hardcode the text and image to avoid the overhead of adding proper image support (via the picker). If the test goes well we can make all this a properly supported part of the Epic component.

https://trello.com/c/xODg9J4r/1217-new-epic-template-with-image

## How to test

In Storybook, there's a new story EpicWithHeaderImage under WorkInProgress/EndOfArticle.

We'll be running a test to validate this new feature.

## Images

Screenshot 2021-09-14 at 15 15 34

## Accessibility

- [ ] [Tested with screen reader](https://github.com/guardian/source/blob/main/docs/06-accessibility.md#screen-readers)
- [ ] [Navigable with keyboard](https://github.com/guardian/source/blob/main/docs/06-accessibility.md#keyboard-navigation)
- [ ] [Colour contrast passed](https://github.com/guardian/source/blob/main/docs/06-accessibility.md#colour-contrast)
- [ ] [The change doesn't use only colour to convey meaning](https://github.com/guardian/source/blob/main/docs/06-accessibility.md#use-of-colour)

### DIFF
--- a/src/BrazeEndOfArticleComponent.ts
+++ b/src/BrazeEndOfArticleComponent.ts
@@ -7,12 +7,13 @@ import {
 } from './NewsletterEpic';
 
 import { COMPONENT_NAME as US_NEWSLETTER_EPIC_NAME, USNewsletterEpic } from './USNewsletterEpic';
-
 import { COMPONENT_NAME as AU_NEWSLETTER_EPIC_NAME, AUNewsletterEpic } from './AUNewsletterEpic';
-
 import { COMPONENT_NAME as UK_NEWSLETTER_EPIC_NAME, UKNewsletterEpic } from './UKNewsletterEpic';
-
 import { COMPONENT_NAME as EPIC_NAME, Epic } from './Epic';
+import {
+    COMPONENT_NAME as EPIC_WITH_HEADER_IMAGE_NAME,
+    EpicWithSpecialHeader,
+} from './EpicWithSpecialHeader';
 import { buildBrazeMessageComponent, ComponentMapping } from './buildBrazeMessageComponent';
 
 type BrazeMessageProps = {
@@ -32,6 +33,7 @@ const END_OF_ARTICLE_MAPPINGS: ComponentMapping<CommonEndOfArticleComponentProps
     [AU_NEWSLETTER_EPIC_NAME]: AUNewsletterEpic,
     [UK_NEWSLETTER_EPIC_NAME]: UKNewsletterEpic,
     [EPIC_NAME]: Epic,
+    [EPIC_WITH_HEADER_IMAGE_NAME]: EpicWithSpecialHeader,
 };
 
 export const BrazeEndOfArticleComponent: React.FC<CommonEndOfArticleComponentProps> =
--- a/src/Epic/index.tsx
+++ b/src/Epic/index.tsx
@@ -3,7 +3,7 @@ import React from 'react';
 import { brand, palette, space } from '@guardian/src-foundations';
 import { ContributionsEpicButtons } from './ContributionsEpicButtons';
 import { body, headline } from '@guardian/src-foundations/typography';
-import { COMPONENT_NAME, canRenderEpic, parseParagraphs } from './canRender';
+import { COMPONENT_NAME, canRender, parseParagraphs } from './canRender';
 export { COMPONENT_NAME };
 import { replaceNonArticleCountPlaceholders } from './placeholders';
 
@@ -23,7 +23,7 @@ const styles = {
         max-width: 620px;
     `,
     epicContainer: css`
-        padding: 4px 8px 12px;
+        padding: ${space[1]}px ${space[2]}px ${space[3]}px;
         border-top: 1px solid #ffe500;
         background-color: #f6f6f6;
         display: flex;
@@ -68,15 +68,17 @@ export type EpicProps = {
     ophanComponentId?: string;
     brazeMessageProps: BrazeMessageProps;
     countryCode?: string;
+    headerSection?: React.ReactNode;
 };
 
 export const Epic: React.FC<EpicProps> = (props: EpicProps) => {
     const {
         brazeMessageProps: { heading, buttonText, buttonUrl, highlightedText },
         countryCode,
+        headerSection,
     } = props;
 
-    if (!canRenderEpic(props.brazeMessageProps)) {
+    if (!canRender(props.brazeMessageProps)) {
         return null;
     }
 
@@ -90,6 +92,8 @@ export const Epic: React.FC<EpicProps> = (props: EpicProps) => {
         <ThemeProvider theme={brand}>
             <div css={styles.epicWrapper}>
                 <section css={styles.epicContainer}>
+                    {headerSection}
+
                     <span css={styles.heading}>{heading}</span>
                     {paragraphs.map((text, index) => (
                         <p key={'paragraph' + index} css={styles.paragraph}>
--- a/src/storybookCommon/argTypes.ts
+++ b/src/storybookCommon/argTypes.ts
@@ -29,3 +29,22 @@ export const ophanComponentIdArgType = {
             '`rc_upsell_test1_variant`.',
     },
 };
+
+export const buildEpicParagraphDocs = (
+    numberOfParagraphs: number,
+): [string, Record<string, unknown>][] =>
+    Array.from({ length: numberOfParagraphs }, (_, idx) => {
+        const name = `paragraph${idx + 1}`;
+        return [
+            name,
+            {
+                name,
+                type: {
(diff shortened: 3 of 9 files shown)


src/EpicWithHeaderImage/index.tsx
```diff
@@ -0,0 +1,91 @@
...
+};
+
+// We use a picture element here with an empty image for < tablet so that we
+// don't load an image unnecessarily when we're not going to render it (which would be
+// the case with an img tag)
+const HeaderSection: React.FC = () => (
+    <div css={headerStyles.container}>
+        <div css={headerStyles.leftContainer}>
+            <span css={headerStyles.text}>
+                You&#8217;re powering open,&nbsp;independent journalism
```
tjmw, line 64: I used a non-breaking space here to keep `open, independent` on the same line. Interested in thoughts on that. It seems to have the desired effect on the browsers/widths I've tested on.

adamnfish: Sounds good if we've checked it doesn't cause problems at any width. A more conservative approach in general is to add word breaks elsewhere. This mean the browser **favours** other splits rather than **forcing** this lack of split, which gives it the flexibility to make this break if doing so is completely necessary.

If this isn't being shown at mobile widths then I imagine this is fine!

src/BrazeEndOfArticleComponent.ts
```diff
@@ -7,12 +7,13 @@ import {
...
-
 import { COMPONENT_NAME as AU_NEWSLETTER_EPIC_NAME, AUNewsletterEpic } from './AUNewsletterEpic';
-
 import { COMPONENT_NAME as UK_NEWSLETTER_EPIC_NAME, UKNewsletterEpic } from './UKNewsletterEpic';
-
 import { COMPONENT_NAME as EPIC_NAME, Epic } from './Epic';
+import {
+    COMPONENT_NAME as EPIC_WITH_HEADER_IMAGE_NAME,
+    EpicWithHeaderImage,
+} from './EpicWithHeaderImage';
```
adamnfish, line 16: Super minor point, but this name kind of doesn't apply anymore! Now it's an epic with a bespoke title header. It does have an image in so this isn't totally misleading! Fine for this test, maybe let's double check our thoughts if we go any further with templating this approach?

src/Epic/index.tsx
```diff
@@ -23,7 +23,7 @@ const styles = {
         max-width: 620px;
     `,
     epicContainer: css`
-        padding: 4px 8px 12px;
+        padding: ${space[1]}px ${space[2]}px ${space[3]}px;
```
adamnfish, line 26: nice improvement

src/Epic/index.tsx
```diff
@@ -90,6 +92,8 @@ export const Epic: React.FC<EpicProps> = (props: EpicProps) => {
         <ThemeProvider theme={brand}>
             <div css={styles.epicWrapper}>
                 <section css={styles.epicContainer}>
+                    {children}
```
adamnfish, line 95: Is there a better name than `children` here? Children is fine and technically accurate, but it doesn't convey the context around how this element is used. This has semantic significance here from where in the markup it is placed that we could convey with this name. Perhaps this is part of a wider interface that I haven't spotted, in which case the specific context here doesn't apply so we shouldn't pick a specific name? If not, perhaps mentioning that this appears as a header would be helpful?

src/EpicWithHeaderImage/index.stories.tsx
```diff
@@ -0,0 +1,99 @@
...
+                componentName={args.componentName}
+                subscribeToNewsletter={() => Promise.resolve()}
+            />
+        </StorybookWrapper>
+    );
+};
+
+export const DefaultStory = StoryTemplate.bind({});
+
+DefaultStory.args = {
```
adamnfish, line 81: Should this default story include header text? II guess it's missing because it's hard-coded in the template, in which case that makes sense.

src/storybookCommon/argTypes.ts
```diff
@@ -29,3 +29,23 @@ export const ophanComponentIdArgType = {
             '`rc_upsell_test1_variant`.',
     },
 };
+
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
```
adamnfish, line 33: Why is this required here? It looks like we should know the types of what we're returning?

adamnfish (approved):
Couple of little Qs, but looks great!

src/BrazeEndOfArticleComponent.ts
tjmw: Ha yeah that's a good point, I think I'd managed to convince myself it was okay because like you say there is still an image in there. I'll give some thought to what a good name would be. I'm hoping this template will be short lived so sticking with the current name doesn't seem _that_ bad, would be nice to do better though.

src/Epic/index.tsx
tjmw: Good questions. The `children` prop is a [React thing](https://reactjs.org/docs/composition-vs-inheritance.html#containment), so if you specify your jsx like this:

```jsx
<A>
  <B />
</A>
```

Then the inner content of `A` is made available to `A` as `children`. But there are other options we could use here to give it a better name. We could capture the arg into a different name using destructuring like this:

```jsx
const A = ({ children: specialHeader ) => { ... }
```

Or alternatively pass the special header like any other prop, like this:

```jsx
<A specialHeader={<B />} />
```

(That last example is how you'd have to do things if you wanted to support a custom header _and_ footer for example).

Any thoughts on which of those might be preferable?

src/storybookCommon/argTypes.ts
tjmw: Good spot, it's not needed (it was left over from a previous iteration of this work).

src/EpicWithHeaderImage/index.stories.tsx
tjmw: Yeah that's correct. The original/core epic template support supports a simple text header which is provided dynamically from Braze. Here we're omitting it as we're using the hardcoded new special header section.

src/EpicWithHeaderImage/index.tsx
tjmw: Also a good point 👍 

> is to add word breaks elsewhere

You mean with `<wbr>`? I've not used those much before, I'll take a look.

I just tried removing the nbsp and the browser is wrapping things exactly the same, maybe this was making a difference in a previous version where this top section was laid out slightly differently. So I'm tempted to remove it. (In a future where that text is dynamically provided we probably wouldn't have that level of control, so maybe better to let the browser get on with it).

adamnfish: Yeah a `<wbr>` tag is one way, IIRC there's an entity for it as well*. It's well-supported in theory but I don't know how strongly browsers would stick to it these days. My experience with this is from a few years back!

* Like me, this page is very old but has some information on the options
    https://www.quirksmode.org/oddsandends/wbr.html

No dramas though, it was fine as-is and even better if it is working without.

src/Epic/index.tsx
adamnfish: It feels a bit different to the React docs there because as well as being "children" it is also **before** other hard-coded children elements (like the `<span>` and the paragraphs. No harm if this makes React sense though, very happy to defer to react expertise!

src/EpicWithHeaderImage/index.tsx
tjmw: Thanks for this! Ahh quirksmode, I haven't thought about that site for ages. I used to refer to it all the time.

src/Epic/index.tsx
tjmw: I've update this, the prop in the `Epic` component is now called `headerSection`. So the new component implementation basically looks like this:

```typescript
export const EpicWithSpecialHeader: React.FC<EpicProps> = (props: EpicProps) => (
    <Epic {...props} headerSection={<HeaderSection />} />
);
```

src/BrazeEndOfArticleComponent.ts
tjmw: I've gone with `EpicWithSpecialHeader` for the name, which I think is a slight improvement on `EpicWithHeaderImage`.

PR #235, opened 2021-09-29
Reduce TTL on cloudfront cache to 1 minute

michaelclapham:
## What does this change?

Reduces the min (and now max) TTL for the cache on the cloudfront used to host braze-components.gutools.co.uk from 1 hour to 1 minute. This reduces the time it takes to confirm that a change is in production storybook which is really helpful when deploying hotfixes. 

https://trello.com/c/LVCDMM2l/1236-revisit-braze-componentsgutoolscouk-caching-strategy

## How to test

Make a change to braze components and check that it shows up in a few minutes as opposed to an hour. Also check the network tab for a changed cache control header.

Changes to braze components show up in a few minutes as opposed to an hour.

This will reduce the use of the cache slightly. Given this is an internal low traffic site the reduced cache hit's shouldn't be a problem.

### DIFF
--- a/cloudformation.yaml
+++ b/cloudformation.yaml
@@ -38,7 +38,8 @@ Resources:
                 DefaultCacheBehavior:
                     AllowedMethods: [HEAD, GET]
                     CachedMethods: [HEAD, GET]
-                    MinTTL: 3600
+                    MinTTL: 60
+                    MaxTTL: 60
                     Compress: true
                     ForwardedValues:
                         QueryString: false


tjmw (approved):
👍  I think this looks like a nice and simple solution for this. I'd been imagining that we might end up doing something more sophisticated, for example bust the cache on deploy, or use etags or similar to communicate to cloudfront when things had changed, but those would definitely be more complex. So +1 to doing this, we can always look into more complex solutions down the line if this doesn't end up being sufficient.

PR #238, opened 2021-10-06
Storybook section dropdown using 75 sections from API

michaelclapham:
## What does this change?

Adds a dropdown / select control to every component in Storybook with all 75 section ids currently returned by the Guardian content API. This allows marketing to select a section id to filter any Braze component to only show when on an article in that section.

## How to test

Launch storybook and check that the control appears below each component.

Section dropdown shows in storybook.

Not easy to copy and paste from dropdown?

## Images

Screenshot 2021-10-06 at 12 02 47

## Accessibility

- [ ] [Tested with screen reader](https://github.com/guardian/source/blob/main/docs/06-accessibility.md#screen-readers)
- [ ] [Navigable with keyboard](https://github.com/guardian/source/blob/main/docs/06-accessibility.md#keyboard-navigation)
- [ ] [Colour contrast passed](https://github.com/guardian/source/blob/main/docs/06-accessibility.md#colour-contrast)
- [ ] [The change doesn't use only colour to convey meaning](https://github.com/guardian/source/blob/main/docs/06-accessibility.md#use-of-colour)


src/storybookCommon/argTypes.ts
```diff
@@ -30,6 +32,16 @@ export const ophanComponentIdArgType = {
     },
 };
 
+export const sectionArgType = {
```
tjmw, line 35: To me these section filters feel like they should have their own category (similar to the `core` category), what do you think? I also reckon maybe this `const` should be called something slightly more generic like `pageContextArgTypes` to make it clear where future filtering controls should go.

tjmw (commented):
This is looking good. One thing I think we could possibly improve in terms of the UI is that all of the other controls we currently have allow you to copy paste over to Braze by someone in marketing. But with a select that isn't possible. It'd be great if we could support that as I think it would reduce friction from the process of specifying the key value pairs in Braze.

michaelclapham:
> This is looking good. One thing I think we could possibly improve in terms of the UI is that all of the other controls we currently have allow you to copy paste over to Braze by someone in marketing. But with a select that isn't possible. It'd be great if we could support that as I think it would reduce friction from the process of specifying the key value pairs in Braze.

I can't find a way of doing this using storybook-controls without modifying storybook itself. Maybe we could write some weird code that would take the value of the dropdown and copy it into another text field, but that feels like it could be quite flaky and confusing to use.

src/storybookCommon/sectionIds.ts
```diff
@@ -0,0 +1,77 @@
+export const guardianSectionIds = [
```
tjmw, line 1: Is it intentional that this list is in here twice under two separate files? :-)

PR #262, opened 2021-12-07
Upgrade to Node 14

tjmw:
## What does this change?

Upgrades from Node 12 -> 14 for building the package and storybook. This is required for the gu-cdk work. Also, Node 12 is EOL April 2022, so we'll want to do this anyway.

### TODO

- [x] Update TeamCity to use Node 14 for this branch (docker image can be configured in build config). (After merging, update all branches to use Node 14.)
- [x] Cut a pre-release from this branch (v5.1.0-0)
- [x] Test prerelease against DCR/frontend.

## Accessibility


package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@guardian/braze-components",
-  "version": "5.0.0",
+  "version": "5.1.0-0",
```
michaelclapham, line 3: Should we also add the engines property to package.json with node 14 in it? { "engines" : { "node" : ">=0.14" } }

tjmw: Hey that's a good question. I'm not sure. 🤔  I guess we don't expect our consumers to use this package in the Node environment, but in the browser. Given that we don't specify it currently, I'm tempted to not make a change in this PR and investigate separately (maybe we have gu wide recommendations for this).

PR #263, opened 2021-12-07
Begin migrating to `@guardian/cdk`

tjmw:
## What does this change?

This begins the migration to `@guardian/cdk`. The first step is to migrate our existing `cloudformation.yaml` by wrapping it with `@guardian/cdk`.

## How to test

Our plan is to use tooling provided by `@guardian/cdk` to compare our current stack with the stack created with CDK, at this stage there shouldn't be any differences.

*Note:* the first commit here upgrades to Node 14, that change actually happens in #262, so we should rebase this PR on main after merging 262.


akash1810 (approved):
AMAZING!

PR #265, opened 2021-12-09
Adopt DNS provisioning into CDK

tjmw:
## What does this change?

Adopt our existing DNS setup into CDK. As we've lined the configuration up exactly, this won't actually cause any changes.

## How to test

We've deployed to `CODE`: https://braze-components.code.dev-gutools.co.uk/


akash1810 (approved):
Wonderful!

PR #269, opened 2021-12-13
End of year banner

michaelclapham:
## Accessibility


src/EndOfYearBanner/index.stories.tsx
```diff
@@ -0,0 +1,74 @@
...
+            description: 'Header text',
+        },
+        body: {
+            name: 'body',
+            type: { name: 'string', required: true },
+            description: 'Body text',
+        },
+        boldText: {
+            name: 'boldText',
+            type: { name: 'string', required: true },
```
tjmw, line 25: We're saying this is required here, but not checking for it in the `canRender` above. I'm not sure which is right (we could check with Gavin) but we should be consistent.

src/EndOfYearBanner/index.tsx
```diff
@@ -0,0 +1,173 @@
...
+const catchAndLogErrors = (description: string, fn: () => void): void => {
+    try {
+        fn();
+    } catch (e) {
+        console.log(`Error (${description}): `, e.message);
+    }
+};
+
+const urlObject = new URL(
+    'https://www.theguardian.com/info/ng-interactive/2020/dec/21/the-guardian-in-2020',
```
tjmw, line 42: I guess let's not merge this until we've got the correct destination sorted!?

tjmw: Actually, should this just be a prop that marketing supply? That's what we do for epics - they provide the full URL with any tracking params through Braze.

src/EndOfYearBanner/index.tsx
```diff
@@ -0,0 +1,173 @@
...
+
+    if (!canRender(props.brazeMessageProps)) {
+        return null;
+    }
+
+    if (!showBanner) {
+        return null;
+    }
+
+    const logToBrazeAndOphan = (internalButtonId: number): void => {
```
tjmw, line 73: Not necessarily as part of this PR, but we should think about abstracting this out. It's common across a few banners I think.

tjmw (commented):
I think this is looking good. I left a few comments/suggestions, see what you think.

src/EndOfYearBanner/index.stories.tsx
michaelclapham: Making it not required just for now.

src/EndOfYearBanner/index.tsx
michaelclapham: Agreed. Changed this so that buttonUrl can and must be passed in.

src/EndOfYearBanner/index.stories.tsx
michaelclapham: I've changed this to no longer be required here or by the can render. I think banner works okay without bold text so making it optional for now seems okay, but you're right we should check with Gavin.

src/EndOfYearBanner/canRender.ts
```diff
@@ -3,9 +3,9 @@ import type { BrazeMessageProps } from './index';
...
  * this means the user won't download the Braze components bundle when the component can't be shown.
  */
-export const COMPONENT_NAME = 'TheGuardianIn2020Banner';
+export const COMPONENT_NAME = 'EndOfYearBanner';
 
 export const canRender = (brazeMessageProps: BrazeMessageProps): boolean => {
-    const { header, body } = brazeMessageProps;
-    return Boolean(header && body);
+    const { header, body, buttonText, buttonUrl } = brazeMessageProps;
+    return Boolean(header && body && buttonText && buttonUrl);
```
tjmw, line 10: Should we be checking here for `imageUrl`? The docs say it's required (I think we said a default didn't make sense?)

tjmw: I believe the `Appbanner` `canRender` also does some checks to verify that the image host etc looks correct, might be worth checking and aligning with how we handle things there.

src/EndOfYearBanner/index.tsx
```diff
@@ -36,26 +40,15 @@ const catchAndLogErrors = (description: string, fn: () => void): void => {
     }
 };
 
-const urlObject = new URL(
-    'https://www.theguardian.com/info/ng-interactive/2020/dec/21/the-guardian-in-2020',
-);
+const DEFAULT_IMAGE =
+    'https://i.guim.co.uk/img/media/c9ba78ef2b1a931aab5ca625ce49646e116b11b3/0_0_3200_1800/3200.png?quality=60&width=930&s=72ff133ea3e4516f5a353213e7a62e8a';
```
tjmw, line 44: I wonder if a default image at this level makes sense, I feel like it might be more confusing for marketing than just saying "you must provide an image". What do you think? If we did want a default an alternative could be to wrap `EndOfYearBanner` in a more specifically named component, which threads the image prop through (I think similar to the way `DigitalSubscriberAppBanner` is implemented on top of `AppBanner`).

src/EndOfYearBanner/index.tsx
```diff
@@ -127,26 +120,21 @@ const TheGuardianIn2020Banner: React.FC<Props> = (props: Props) => {
                     <p css={styles.paragraph}>
                         {body}
                         <br />
```
tjmw, line 122: Should the `<br/>` be included conditionally only if the `boldText` prop is supplied?

src/EndOfYearBanner/index.tsx
tjmw: (This is a genuine question btw! I'm not sure how I feel about the default! Maybe Yusuf will have an opinion.)

src/EndOfYearBanner/index.stories.tsx
```diff
@@ -4,10 +4,17 @@ import { StorybookWrapper } from '../utils/StorybookWrapper';
 import { knobsData } from '../utils/knobsData';
 import { coreArgTypes } from '../storybookCommon/argTypes';
 import { BrazeMessageProps } from '.';
+import { grid, withGrid } from '../../.storybook/grid/withGrid';
 
 export default {
-    component: 'TheGuardianIn2020Banner',
-    title: 'Banner/TheGuardianIn2020Banner',
+    component: 'EndOfYearBanner',
+    title: 'Banner/EndOfYearBanner',
```
tjmw, line 11: I'd say maybe this should be in the work-in-progress folder until this PR is merged, a new `@guardian/braze-components` version is released and is available on frontend/DCR. What do you reckon? I think that's how we've done things in the past (we should add this to the README if we agree that flow makes sense).

tjmw (commented):
This is looking good! Nice work getting the image picking functionality there, I wasn't sure how straightforward onboarding a second component to use it would be.

I've left a few comments and questions, see what you think.

One note is that I'm not sure we've ever actually used `AppBanner` (the only other place we use the image picker) in production before on a real campaign. I think the first few times marketing set up an in-app message using the image picker (so this `EndOfYearBanner` component, or `AppBanner`) we should inspect the i.guim image URL which is created for them to verify the size/quality etc looks appropriate.

src/EndOfYearBanner/index.tsx
michaelclapham: I've removed the default image as the imageUrl is now required by canRender and the story.

src/EndOfYearBanner/index.stories.tsx
```diff
@@ -17,6 +17,7 @@ export default {
     },
     argTypes: {
         ...coreArgTypes,
+        ...ophanComponentIdArgType,
```
tjmw, line 20: Nice! This reminded me of a thing I was thinking about earlier but forgot to mention. If this is required now we should make sure the `canRender` for this banner reflects that (i.e. `canRender` should return `false` if the `ophanComponentId` is missing.

src/EndOfYearBanner/canRender.ts
michaelclapham: Yep I agree. I've added this i.guim check to this banners canRender now.

src/EndOfYearBanner/index.tsx
michaelclapham: Yep this is done now.

src/EndOfYearBanner/index.stories.tsx
michaelclapham: Yep this is now in WorkInProgress here: 
https://github.com/guardian/braze-components/blob/end-of-year-banner/src/BannerWithLink/index.stories.tsx#L11

src/EndOfYearBanner/index.stories.tsx
tjmw: I just updated the `canRender` in 2c74bd7 to reflect this. Does that sound okay @michaelclapham?

tjmw (approved):

PR #279, opened 2022-01-11
Adopt existing Cloudfront Distribution

tjmw:
## What does this change?

Adopts the existing CloudFront distribution in our CDK so we can migrate to CDK with no downtime (we want to keep the custom domain, and migrating cloudfront dist aliases is complicated).

This also means we no longer need the load the old/existing cloudformation template as everything is now handled by CDK, so the cloudformation has been deleted.

## How to test

Visit https://braze-components.code.dev-gutools.co.uk/

## Accessibility


akash1810 (approved):

akash1810:
This is fantastic! Great work!

PR #282, opened 2022-01-17
BannerWithLink style tweaks

tjmw:
## What does this change?

This PR updates the `BannerWithLink` colours to the designs from marketing for the Gu in 2021 moment.

I made a couple of simplifications along the way to accommodate the new colours, for example:

* Removed the hover effect on the close button and main link button
* Simplified the info icon to be a single colour with transparent background

## How to test

The changes are deployed to the [CODE storybook](https://braze-components.code.dev-gutools.co.uk/?path=/story/workinprogress-banner-bannerwithlink--default-story).

## Images

Screenshot 2022-01-17 at 12 10 30

Screenshot 2022-01-17 at 12 11 34

## Accessibility

- [x] [Navigable with keyboard](https://guardian.github.io/source/?path=/docs/docs-06-accessibility--page#keyboard-navigation)
- [x] [Colour contrast passed](https://guardian.github.io/source/?path=/docs/docs-06-accessibility--page#colour-contrast)

### DIFF
--- a/src/BannerWithLink/index.stories.tsx
+++ b/src/BannerWithLink/index.stories.tsx
@@ -54,7 +54,7 @@ export default {
 
 const StoryTemplate = (args: BrazeMessageProps & { componentName: string }): ReactElement => {
     const imageUrl = grid(
-        'https://i.guim.co.uk/img/media/c9ba78ef2b1a931aab5ca625ce49646e116b11b3/0_0_3200_1800/3200.png?quality=60&width=930&s=72ff133ea3e4516f5a353213e7a62e8a',
+        'https://i.guim.co.uk/img/media/3bf8e0a7b3d59ba66ac91792ceed73a7b0536f62/0_0_931_523/master/931.png?width=930&quality=45&auto=format&s=629e348cc332f79b8ae9beda5c3f3429',
     );
 
     const brazeMessageProps: BrazeMessageProps = {
--- a/src/BannerWithLink/styles.ts
+++ b/src/BannerWithLink/styles.ts
@@ -1,7 +1,5 @@
 import { css } from '@emotion/react';
-import { from, until, body, neutral, space } from '@guardian/source-foundations';
-
-const bodyColor = '#666';
+import { from, until, space } from '@guardian/source-foundations';
 
 export const styles = {
     image: css`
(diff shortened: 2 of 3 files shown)


src/BannerWithLink/index.stories.tsx
```diff
@@ -54,7 +54,7 @@ export default {
 
 const StoryTemplate = (args: BrazeMessageProps & { componentName: string }): ReactElement => {
     const imageUrl = grid(
-        'https://i.guim.co.uk/img/media/c9ba78ef2b1a931aab5ca625ce49646e116b11b3/0_0_3200_1800/3200.png?quality=60&width=930&s=72ff133ea3e4516f5a353213e7a62e8a',
+        'https://i.guim.co.uk/img/media/3bf8e0a7b3d59ba66ac91792ceed73a7b0536f62/0_0_931_523/master/931.png?width=930&quality=45&auto=format&s=629e348cc332f79b8ae9beda5c3f3429',
```
tjmw, line 57: This just updates the _default_ image used by the story for this banner.

src/BannerWithLink/styles.ts
```diff
@@ -1,9 +1,14 @@
 import { css } from '@emotion/react';
-import { from, until, body, neutral, space } from '@guardian/source-foundations';
+import { from, until, space, neutral } from '@guardian/source-foundations';
+import { ButtonTheme, buttonThemeReaderRevenueBrandAlt } from '@guardian/source-react-components';
 
-const bodyColor = '#666';
+const backgroundColor = '#f79e1b';
```
michaelclapham, line 5: I wonder if we should name this component OrangeBannerWithLink so that any campaigns that want to use this orange background don't have to be updated if BannerWithLink wants to be used for a different coloured campaign in the future.

src/BannerWithLink/styles.ts
```diff
@@ -52,16 +46,32 @@ export const styles = {
...
+            background: transparent;
+        }
+    `,
+    heading: css`
+        color: ${neutral[0]};
+    `,
+};
+
+// Use this theme but override the colours to match the moment colours
+export const overrridenReaderRevenueTheme: { button: ButtonTheme } = {
```
michaelclapham, line 71: I remember this was used to make a button not change colour when depending on source for the first time (or a source upgrade or something similar). Should we create a ticket to follow up on removing this code? Eventually the palette that marketing want to use for Braze components should live in Source, instead of us making little tweaks to the existing palettes.

src/BannerWithLink/styles.ts
tjmw: Good question! I'm tempted to leave this as-is for now while we wait and see how we want to handle palettes in banners. I think it's probably low risk in terms of having to update canvases as this look back "moment" will have a finite lifespan.

src/BannerWithLink/styles.ts
tjmw: Yeah agreed, hopefully this can be removed once we figure out how to handle palettes for the banner. (Just to note: this isn't an addition in this PR, it just moved from the main component index file). Though you pointing this out does make me realise there's a tiny typo in the const name here so will fix that!

michaelclapham (approved):

PR #304, opened 2022-03-11
Epic Remind Me Button Tracking

tjmw:
## What does this change?

This PR builds on #303 by wiring up the tracking on the Remind Me button (which is essential, so that we can know to schedule the email in Braze when a user clicks).

I was finding that the detail of all of our underlying components knowing about `submitComponentEvent` (Ophan) and the Braze tracking callback, which are both provided by the platform, a bit unwieldy. The underlying components always just trigger both, and that logic was duplicated in multiple places. More importantly, I don't think the underlying components should care about this level of detail.

So this PR also includes a refactoring where the higher-level entry component for each slot (`BrazeEndOfArticleComponent` and `BrazeBannerComponent`) takes the two click handling functions and combines them into a single callback `trackClick` which is passed to the individual components. So far I've only made us of this for epic components, but we can follow up in another PR to do this for banners.

## Accessibility

### DIFF
--- a/src/Epic/ContributionsEpicButtons.tsx
+++ b/src/Epic/ContributionsEpicButtons.tsx
@@ -5,6 +5,9 @@ import { Button, LinkButton, SvgArrowRightStraight } from '@guardian/source-reac
 
 import { RemindMeConfirmation } from './RemindMeConfirmation';
 
+const PRIMARY_BUTTON_INTERNAL_ID = 0;
+const REMIND_ME_BUTTON_INTERNAL_ID = 1;
+
 const buttonWrapperStyles = css`
     margin: ${space[4]}px ${space[2]}px ${space[1]}px 0;
     display: flex;
@@ -53,9 +56,10 @@ const remindMeButtonOverrides = css`
 interface PrimaryButtonProps {
     buttonText: string;
     buttonUrl: string;
+    trackClick: (buttonId: number) => void;
 }
 
-const PrimaryButton = ({ buttonUrl, buttonText }: PrimaryButtonProps) => (
+const PrimaryButton = ({ buttonUrl, buttonText, trackClick }: PrimaryButtonProps) => (
     <div css={buttonMargins}>
         <ThemeProvider theme={contributionsTheme}>
             <LinkButton
@@ -65,6 +69,7 @@ const PrimaryButton = ({ buttonUrl, buttonText }: PrimaryButtonProps) => (
                 target="_blank"
                 rel="noopener noreferrer"
                 priority={'primary'}
+                onClick={() => trackClick(PRIMARY_BUTTON_INTERNAL_ID)}
             >
                 {buttonText}
             </LinkButton>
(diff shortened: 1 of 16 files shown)


src/Epic/ContributionsEpicButtons.tsx
```diff
@@ -65,6 +66,7 @@ const PrimaryButton = ({ buttonUrl, buttonText }: PrimaryButtonProps) => (
                 target="_blank"
                 rel="noopener noreferrer"
                 priority={'primary'}
+                onClick={() => trackClick(0)}
```
arausuy, line 69: I'm not super sure what the magic zero is here

arausuy: Its the button ID? (just looking bellow)

src/buildBrazeMessageComponent.tsx
```diff
@@ -1,15 +1,27 @@
...
+    trackClick: TrackClick;
+}
+
 export type ComponentMapping<A> = {
     [key: string]: React.FC<A>;
 };
 
-export function buildBrazeMessageComponent<A extends HasComponentName>(
-    mappings: ComponentMapping<A>,
+export function buildBrazeMessageComponent<A extends HasComponentName & HasClickTracking>(
```
arausuy, line 22: I'm guessing the ampersand is either an intersection type? i.e A must implement both `HasComponentName` and `HasClickTracking`

src/Epic/ContributionsEpicButtons.tsx
tjmw: Yeah exactly. I think pulling this out to a const to give it a name would be helpful to improve clarify, I'll do that.

src/buildBrazeMessageComponent.tsx
```diff
@@ -1,15 +1,27 @@
 import React from 'react';
+import type { BrazeClickHandler, SubmitComponentEvent, TrackClick } from './utils/tracking';
+import { buildTrackClick, OphanComponentType } from './utils/tracking';
 
 interface HasComponentName {
     componentName: string;
 }
 
+interface HasClickTracking {
```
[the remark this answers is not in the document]
tjmw: Good point, yeah I can see now that these aren't the clearest. We're saying at the type level that we're able to go from some type with the `HasClickTracking` interface (i.e. has `submitComponentEvent` and `logButtonClickWithBraze` methods) to a type with the `HasTrackClick` interface (i.e. provides a single unified `trackClick` method). I'll have a think about better names.

src/buildBrazeMessageComponent.tsx
tjmw: Yep that's right 👍  https://www.typescriptlang.org/docs/handbook/2/objects.html#intersection-types

src/buildBrazeMessageComponent.tsx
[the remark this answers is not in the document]
tjmw: I've tried to improve the naming here (and also added a clarifying comment).

src/Epic/ContributionsEpicButtons.tsx
tjmw: Have pulled these out to consts 👍

arausuy (approved):

PR #310, opened 2022-03-24
Changed colour theme for the BannerWithLink

arausuy:
## What does this change?

This change removes the orange colouring used for the Moment we were previously supporting. The colours are now defaulting to our standard grey.

## Accessibility

### DIFF
--- a/src/BannerWithLink/index.tsx
+++ b/src/BannerWithLink/index.tsx
@@ -1,17 +1,9 @@
 import React, { useState } from 'react';
-import { ThemeProvider } from '@emotion/react';
-import {
-    Button,
-    LinkButton,
-    SvgCross,
-    SvgInfo,
-    buttonThemeReaderRevenueBrandAlt,
-} from '@guardian/source-react-components';
+import { Button, LinkButton, SvgCross, SvgInfo } from '@guardian/source-react-components';
 import { OphanComponentEvent } from '@guardian/libs';
 
 import { BrazeClickHandler } from '../utils/tracking';
-import { styles as commonStyles } from '../styles/bannerCommon';
-import { overridenReaderRevenueTheme, styles } from './styles';
+import { styles } from '../styles/bannerCommon';
 import { canRender, COMPONENT_NAME } from './canRender';
 export { COMPONENT_NAME };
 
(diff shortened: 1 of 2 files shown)


arausuy:
I've gone for just stripping the original different css out as a first pass, partly so I could leverage chromatic to see if the inclusion of the image css would alter the rendering of any of the other banners.

While this approach seems to work I think there could be a more elegant one that would enable us to have multiple colour themes but also not to just delete the alternative reader revenue palette as these moments may occur soon.

arausuy:
@tjmw I wasn't sure on the best approach here, my inexperience shines through! In my mind it is fine to just slightly adjust our current banners to have a single piece of css styling the images but I'm more than happy to be wrong and have a separate set of statements for the `BannerWithLink`!

tjmw:
@arausuy Yeah I think that sounds fine 👍  I think the reason I took the approach of having specific styles for `BannerWithLink` was just that there was time pressure and I wanted to apply changes and not worry about the other banners. I think the diffs in Chromatic look fine.

Looking at Chromatic made me realise that it's probably time to remove the `SpecialEditionBanner` - it's not been used in a while so not serving any purpose other than bloating the code (not in this PR necessarily).

I noticed that the body text size on `BannerWithLink` has increased to be inline with the other banners which means that at desktop width there's quite a lot of empty space with the default image/text in the story (more than before):

Screenshot 2022-03-29 at 10 46 15

It's a really tricky thing to get right though, a different image/text would give different results. Maybe at some point getting a designer to look at the banners more holistically would help us get the right balance.

src/BannerWithLink/index.stories.tsx
```diff
@@ -8,7 +8,7 @@ import { grid, withGrid } from '../../.storybook/grid/withGrid';
 
 export default {
     component: 'BannerWithLink',
-    title: 'Banner/BannerWithLink',
+    title: 'WorkInProgress/Banner/BannerWithLink',
```
arausuy, line 11: @tjmw I think this is all it takes to move things to WIP (Just borrowing the approach from a previous [PR](https://github.com/guardian/braze-components/pull/284))

tjmw: Yeah, that looks right to me! 👍

tjmw (approved):

PR #313, opened 2022-03-29
Remove Special Edition Banner as not in use anymore

arausuy:
## What does this change?

We haven't had any campaigns use our special edition banner for quite a while, we see no impressions for this banner in the lake and can remove this to reduce bundle size. 

## How to test

Run storybook and see that it is no longer available.

## Accessibility


tjmw (approved):
👍  This looks great!

There are some changes according to Chromatic, but I think that's because we've got un-accepted changes on main from the previous work you did, so the baseline hasn't been updated. I think if we accept the Chromatic changes on main, the diffs reported on this PR will go away.

PR #316, opened 2022-04-06
Content cards spike

adamnfish:
## What does this change?

BrazeCards is a mechanism for providing persistent 'notification' messages, in contrast to the existing single-impression (In-App-Message backed) BrazeMessages behaviour.

This PR introduces the basic abstraction for content cards, without providing any specific slot component. BrazeCards looks very similar to BrazeMessages, but exposes a different interface.

- it returns collections of cards for the given slot, instead of selecting a single message
- Braze's SDK serves cards out its own cache, so there's no need to return a `promise`
- the above means we can also skip our own cache implementation

This change also splits the representation of 'slots' into `MessageSlot`s and `CardSlot`s. This change encodes in the types the distinction between slots which will display the single highest-priority message (if available), versus those that are able to prompt the user on multiple counts simultaneously.

I expect in the future we'd consider setting up notification slots to support multiple message streams of differeing priority, but this is something we should add as we develop a better intuition of how this feature will be used. Well-considered [prior art](https://developer.android.com/training/notify-user/channels) exists for these features, when the need arises.

## How to test

This draft includes tests, but the next step is to do some work exploring the DCR integration.

We won't be able to say this is a success until we have created a real slot on the page and started running tests using this card mechanism.

This will slightly increase the bundle size of this library's 'core'. Other risks are minimal for now, before we've integrated this with a real slot on the site.

## Images

N/A

## Accessibility

N/A

### DIFF
--- a/src/logic/BrazeCards.test.ts
+++ b/src/logic/BrazeCards.test.ts
@@ -1,0 +1,247 @@
+import fc from 'fast-check';
+import appboy from '@braze/web-sdk-core';
+import { BrazeCard, BrazeCards } from './BrazeCards';
+import { CardSlotName, Extras } from './types';
+
+function errorHandler(error: Error, identifier: string): void {
+    console.log(identifier, error);
+}
+
+describe('braze content cards', () => {
+    describe('BrazeCards', () => {
+        describe('getCardsForProfileBadge', () => {
+            test('returns empty array if there are no content cards for this user', () => {
+                jest.spyOn(appboy, 'getCachedContentCards').mockImplementation(() => {
+                    return new appboy.ContentCards([], new Date());
+                });
+                const brazeCards = new BrazeCards(appboy, errorHandler);
+                const result = brazeCards.getCardsForProfileBadge();
+                expect(result).toEqual([]);
+            });
+
+            test("returns BrazeCard instances for each of the profile badge cards in the user's cached card feed", () => {
+                jest.spyOn(appboy, 'getCachedContentCards').mockImplementation(() => {
+                    return new appboy.ContentCards(
+                        [
+                            testCard('ProfileBadge', 'card-id-1', { field1: 'value-1' }),
+                            testCard('ProfileBadge', 'card-id-2', { field1: 'another-value' }),
+                            testCard('ProfileBadge', 'card-id-3', { field2: 'yet-another-value' }),
+                        ],
+                        new Date(),
+                    );
+                });
+                const brazeCards = new BrazeCards(appboy, errorHandler);
+                const result = brazeCards.getCardsForProfileBadge();
+                const cardIds = result.map((card) => card.id);
+                expect(cardIds).toEqual(['card-id-1', 'card-id-2', 'card-id-3']);
+            });
+
+            test("uses Braze's card ID", () => {
+                fc.assert(
+                    fc.property(fc.string(), (id) => {
+                        jest.spyOn(appboy, 'getCachedContentCards').mockImplementation(() => {
+                            return new appboy.ContentCards(
+                                [testCard('ProfileBadge', id, {})],
+                                new Date(),
+                            );
+                        });
+                        const brazeCards = new BrazeCards(appboy, errorHandler);
+                        const result = brazeCards.getCardsForProfileBadge();
+                        expect(result[0].id).toEqual(id);
+                    }),
+                );
+            });
+
+            test('uses the extras record that Braze provides', () => {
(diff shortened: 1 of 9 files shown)


src/logic/BrazeCards.test.ts
```diff
@@ -0,0 +1,247 @@
...
+import { CardSlotName, Extras } from './types';
+
+function errorHandler(error: Error, identifier: string): void {
+    console.log(identifier, error);
+}
+
+describe('braze content cards', () => {
+    describe('BrazeCards', () => {
+        describe('getCardsForProfileBadge', () => {
+            test('returns empty array if there are no content cards for this user', () => {
```
tjmw, line 13:
```suggestion
            it('returns empty array if there are no content cards for this user', () => {
```

Super minor and nit-picky, but we've tended to use `it` for these assertion blocks elsewhere in the codebase, if we want to be consistent here (I also like the way the entire line reads with `it`).

src/logic/BrazeCards.test.ts
```diff
@@ -0,0 +1,247 @@
...
+                jest.spyOn(appboy, 'getCachedContentCards').mockImplementation(() => {
+                    return new appboy.ContentCards([], new Date());
+                });
+                const brazeCards = new BrazeCards(appboy, errorHandler);
+                const result = brazeCards.getCardsForProfileBadge();
```
tjmw, line 18:
```suggestion

                const result = brazeCards.getCardsForProfileBadge();
                
```

Again also nitpicky and my personal preference (so feel free to ignore!) but I try to use the Arrange Act Assert pattern in these tests as much as possible (this line being the Act portion), with a newline separating each phase. I find it helps me visualise the different parts of the test.

src/logic/BrazeCards.test.ts
```diff
@@ -0,0 +1,247 @@
...
+                expect(result).toEqual([]);
+            });
+
+            test("returns BrazeCard instances for each of the profile badge cards in the user's cached card feed", () => {
+                jest.spyOn(appboy, 'getCachedContentCards').mockImplementation(() => {
+                    return new appboy.ContentCards(
+                        [
+                            testCard('ProfileBadge', 'card-id-1', { field1: 'value-1' }),
```
tjmw, line 26: 👍  for creating this helper method 🎉

src/logic/BrazeCards.test.ts
```diff
@@ -0,0 +1,247 @@
...
+                        });
+                        const brazeCards = new BrazeCards(appboy, errorHandler);
+                        const result = brazeCards.getCardsForProfileBadge();
+                        expect(result[0].extras).toEqual(extras);
+                    }),
+                );
+            });
+
+            test('returns no cards if the cached feed contains only unsupported cards', () => {
+                // cards with no extras or missing the slotName, so they cannot be handled by this library
```
tjmw, line 74: Do we want to include the "no extras" case in the test below, to ensure it gets filtered out correctly? Related, should this comment be moved into the test description? It's more specific than the "unsupported cards" description we've currently got. So something like:

```
test('returns no cards if the cached feed contains only cards with no extras or missing the slotName', () => {
```

src/logic/BrazeCards.test.ts
```diff
@@ -0,0 +1,247 @@
...
+                'ProfileBadge',
+                appboyCard,
+                appboy,
+                mockErrorHandler,
+            );
+
+            brazeCard.logImpression();
+
+            expect(mockAppboyFn).toHaveBeenCalledWith([appboyCard]);
+            expect(mockErrorHandler).toHaveBeenCalledTimes(0);
```
tjmw, line 132: Super nitpicky, but I think you could also write this as:

```
expect(mockErrorHandler).not.toHaveBeenCalled();
```

which I think reads _marginally_ nicer.

src/logic/BrazeCards.ts
```diff
@@ -0,0 +1,152 @@
...
+
+    /**
+     * Returns the card's key/value pairs.
+     *
+     * We know this can't be empty because there must have been at least a `slotName`
+     * field present to be able to generate the card.
+     */
+    get extras(): Extras {
+        const data = this.card.extras;
+        // since empty extras is impossible, let's add this case to eliminate undefined from the type
```
tjmw, line 83: 👍

tjmw (approved):
🏅  This looks great, nice work. I've left a bunch of comments but they're all pretty minor and nitpicky. See what you think, happy to chat over any of the feedback if anything doesn't make sense (or you don't agree!)

PR #342, opened 2022-06-06
Upgrade json-schema to 0.4.0

tjmw:
## What does this change?

There is a vulnerability in earlier versions so upgrade. We were pulling in json-schema as a transitive dependency of a couple of other libraries we're using (chromatic and also rollup-plugin-filesize) which is why those libs have been upgraded.

## Accessibility


ecguardian (approved):
LGTM 👍

PR #362, opened 2022-09-19
Add support for multiple section filters

tjmw:
## What does this change?

We've supported specifying a section key/value pair in the message configuration in Braze for a while. Up until now there was a limitation that you could only provide a single section, as the comparison with the page section was a simple equality check.

This adds support for specifying multiple sections separated by the `|` (pipe) character. For example, if you wanted your message to appear on environment OR science articles you would supply a key/value pair like this:

"section" -> "environment|science"

https://trello.com/c/PUIIGlVX/735-braze-web-messaging-support-targeting-multiple-sections

## Accessibility

-   [ ] [Tested with screen reader](https://github.com/guardian/accessibility/blob/main/people-and-technology/03-visual.md#screen-reader)
-   [ ] [Navigable with keyboard](https://github.com/guardian/accessibility/blob/main/people-and-technology/02-physical.md#keyboard)
-   [ ] [Colour contrast passed](https://github.com/guardian/accessibility/blob/main/people-and-technology/03-visual.md#contrast)
-   [ ] [The change doesn't use only colour to convey meaning](https://github.com/guardian/accessibility/blob/main/people-and-technology/03-visual.md#use-of-colour)

PR #363, opened 2022-09-22
Storybook Multi-section Filter

tjmw:
## What does this change?

In #362 we added support for specifying that a message should appear on multiple sections (previously there was a limitation to a single section). This PR updates the storybook `Article Context Filters` tab to make it clear multiple sections are supported. When multiple sections are selected a pipe (`|`) separated list is generated which can be copy pasted into Braze.

In future it might be nice to use a fancier multi-select component, but for now I'm keeping it simple and going with the basic HTML version.


tomrf1 (approved):
👍

PR #367, opened 2022-10-06
Upgrade @guardian/cdk to v48.0.0

tjmw:
## What does this change?

Upgrades `@guardian/cdk` to v48.0.0. Quite a bit has changed since we introduced `@guardian/cdk` so there are some changes to our CDK implementation. I think the diffs in the snapshot make sense.


tomrf1 (approved):
👍

PR #373, opened 2022-11-10
Epic - hide payment icons

KaliedaRik:
## What does this change?
Adds the ability to suppress the payment icons beneath the primary CTA by setting a new attribute

### Screenshots

**Epic (classic)**
Screenshot 2022-11-10 at 14 19 34

Screenshot 2022-11-10 at 14 19 48

**Epic with reminder**
Screenshot 2022-11-10 at 14 20 11

Screenshot 2022-11-10 at 14 20 24

**Epic with special header**
Screenshot 2022-11-10 at 14 20 45

Screenshot 2022-11-10 at 14 20 59


src/Epic/ContributionsEpicButtons.tsx
```diff
@@ -123,8 +124,10 @@ export const ContributionsEpicButtons = ({
     remindMeConfirmationText,
     remindMeConfirmationHeaderText,
     trackClick,
+    hidePaymentIcons,
 }: ContributionsEpicButtonsProps): JSX.Element => {
     const [sectionState, setSectionState] = useState<SectionState>('DEFAULT');
+    const showPaymentIcons = hidePaymentIcons != null && hidePaymentIcons === 'true' ? false : true;
```
tjmw, line 130: I wonder, can this be simplified to:

`const showPaymentIcons = !(hidePaymentIcons === 'true');`

? I.e. the only time they should be hidden is if `hidePaymentIcons` equals the string `'true'` and in all other cases show them, we we flip the result of compaing to `'true'`.

tjmw (approved):
Looks great! One minor question/suggestion on the logic.

src/Epic/ContributionsEpicButtons.tsx
tomrf1, line 130: is this equivalent to `const showPaymentIcons = hidePaymentIcons !== 'true'`?

tjmw: Ha I made (almost) the exact same suggestion above :-) This is better though!

KaliedaRik: Fixed with TomF suggested alternative

KaliedaRik: Fixed using this alternative code

PR #375, opened 2022-11-17
Fix margin CSS on epic body paragraphs

KaliedaRik:
## What does this change?
A small styling fix to prevent short Epic body paragraphs centering themselves when they should be left-aligned

## How to test
Tested in Storybook by deliberately shortening paragraph copy.

The bug only effected default Epic components, not any of the other epics (verified in Storybook)

## Screenshots

**Before fix**
Screenshot 2022-11-17 at 15 37 05

**After fix**
Screenshot 2022-11-17 at 15 37 45

### DIFF
--- a/src/Epic/index.tsx
+++ b/src/Epic/index.tsx
@@ -30,7 +30,7 @@ const styles = {
         flex-direction: column;
     `,
     paragraph: css`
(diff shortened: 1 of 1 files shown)


tomrf1 (approved):
nice

PR #379, opened 2022-12-01
Log impressions as content cards

tjmw:
## What does this change?

The `logCardImpressions` method takes an optional second boolean argument. This determines whether to log the impression as a content card event (as opposed to a legacy news feed event, which is apparently a predecessor to content cards).


KaliedaRik (approved):
LGTM

PR #380, opened 2022-12-08
Bump `@guardian/source-react-components-development-kitchen` and `@guardian/libs`

tjmw:
## What does this change?

We rely on both `@guardian/source-react-components-development-kitchen` and `@guardian/libs` as peer dependencies, provided by the web platforms.

This PR loosens the peer dependencies for both and updates the dev dependency to the latest version.

### DIFF
--- a/package.json
+++ b/package.json
@@ -97,7 +97,7 @@
     "@guardian/libs": "^3.1.0",
     "@guardian/source-foundations": "^7.0.1",
     "@guardian/source-react-components": "^9.0.0",
-    "@guardian/source-react-components-development-kitchen": "^7.1.1",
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -89,15 +89,16 @@
     "rollup-plugin-visualizer": "^4.1.1",
     "storybook": "^6.5.4",
     "ts-jest": "^28.0.3",
+    "tslib": "^2.4.1",
```
tjmw, line 92: This is a peer dep of latest `@guardian/libs`.

KaliedaRik (approved):
LGTM

PR #385, opened 2023-01-27
New newsletter Epic templates: AUS-Afternoon and the-Grid

KaliedaRik:
## What does this change?
Creates two new Epic newsletter templates:

Screenshot 2023-02-03 at 14 46 31

Screenshot 2023-02-03 at 14 46 44


src/BrazeEndOfArticleComponent.ts
```diff
@@ -1,29 +1,46 @@
...
-import { COMPONENT_NAME as EPIC_NAME, Epic } from './Epic';
+
+// New name Newsletter Epics
 import {
-    COMPONENT_NAME as EPIC_WITH_HEADER_IMAGE_NAME,
-    EpicWithSpecialHeader,
-} from './EpicWithSpecialHeader';
+    COMPONENT_NAME as EPICNEWSLETTER_AU_AFTERNOONUPDATE_NAME,
+    EpicNewsletter_AU_AfternoonUpdate,
+} from './EpicNewsletter_AU_AfternoonUpdate';
```
tjmw, line 39: No big deal either way, but I wonder if it'd be nice to be consistent with naming? Could this be `AusAfternoonUpdateNewsletterEpic` for example (and `TheGuideNewsletterEpic` for the other one)? (to have a common suffix with the existing components)

src/BrazeEndOfArticleComponent.ts
```diff
@@ -1,29 +1,46 @@
...
+// Common Epic imports
+import {
+    buildBrazeMessageComponent,
+    ComponentMapping,
+    HasConsolidatedTrackClick,
+} from './buildBrazeMessageComponent';
+import type { BrazeClickHandler, SubmitComponentEvent } from './utils/tracking';
+
+// Why are we mapping this component?
+// Isn't it the blueprint on top of which other newsletter epics get built?
```
tjmw, line 12: This is a good question. Yep you're right it's the blueprint, but this can be the generic version which marketing would like. I think the main reason it's not usable already is related to tooling (for example making selecting an image really easy - see the embedded grid image picker approach used by the `BannerWithLink` component)

KaliedaRik: Good thought for the upcoming work, though I suggest we create a new folder for the new self-serve template so existing canvases/campaigns don't break when we develop/deploy it. Then Marketing can migrate canvases at their leisure.

src/BrazeEndOfArticleComponent.ts
KaliedaRik: My personal preference is for semantic naming - major_minor_whatever - which I think should be EpicNewsletter_REGION-if-applicable_newsletter-name ... if that makes sense?

It's a moot thought though, given that we'll shortly be moving to a more self-serve template so we don't have to build these bespoke templates anymore

PR #386, opened 2023-01-31
Move CI from TeamCity to GitHub Actions

KaliedaRik:
## What does this change?
It does what it says in the title.


tomrf1 (approved):
nice!

PR #387, opened 2023-02-02
Update peer deps to reflect what's on the platforms

tjmw:
## What does this change?

Updates the peer dependency versions required by this package from the consumers (web platforms). Also update the dev deps to the reflect this.

This is to reflect the versions which are already being provided by the web platforms.

## Accessibility


package.json
```diff
@@ -42,11 +42,11 @@
     "@emotion/react": "^11.1.2",
     "@guardian/eslint-plugin-source-react-components": "^9.0.0",
```
arelra, line 43: Can we bump `@guardian/eslint-plugin-source-react-components` too?

package.json
```diff
@@ -42,11 +42,11 @@
...
     "@guardian/grid-client": "^1.1.1",
-    "@guardian/libs": "^10.1.1",
+    "@guardian/libs": "^13.1.0",
     "@guardian/node-riffraff-artifact": "^0.3.2",
     "@guardian/source-foundations": "^7.0.1",
```
arelra, line 47: Can we bump `@guardian/source-foundations` too?

package.json
```diff
@@ -95,10 +95,10 @@
   },
   "peerDependencies": {
     "@emotion/react": "^11.1.2",
-    "@guardian/libs": "^7.1.4 || ^10.1.1",
+    "@guardian/libs": "^7.1.4 || ^10.1.1 || ^12.0.1 || ^13.1.0",
     "@guardian/source-foundations": "^7.0.1",
-    "@guardian/source-react-components": "^9.0.0",
-    "@guardian/source-react-components-development-kitchen": "^6.0.2 || ^7.1.1",
+    "@guardian/source-react-components": "^9.0.0 || ^11.2.0",
+    "@guardian/source-react-components-development-kitchen": "^6.0.2 || ^7.1.1 || ^8.2.0 || ^9.2.0",
```
arelra, line 101: If we do a major release does that mean we can drop these ranges and simply do a `^` semver of each dependency at its latest version? 🤔

arelra: Then when we bump `braze-components` in DCR and Frontend we bump all the associated peer deps too which should allow better alignment?

Pinging @mxdvl for comments

mxdvl: Agreed that it may be advisable to do a major release soon and bump the lowest requirements. [You should always pin your `devDependencies` to the lowest possible match](https://github.com/guardian/recommendations/blob/main/npm-packages.md#pin-the-lowest-possible-version-of-peerdependencies-in-your-packages-devdependencies). For example, that would be `@guardian/libs@7.1.4`.

I would also use wider ranges, like for example any version from 6.0.2 and 9.x.x : `>=6.02  <10`.

package.json
tjmw: Yep definitely can do. It looks like both DCR and frontend are on v7 so I guess we'd want to upgrade them at the same time?

package.json
tjmw: Yeah agree that we could lose some of the lower dependencies, I'll do that.

> You should always pin your devDependencies to the lowest possible match

I'm really interested in this bit. It seems that there's no ideal solution here, if your peer deps follow semver and you specify multiple major releases in your peer deps then which ever dev dependency you choose, that's what your tests will run against, and there's always a danger there are backwards incompatible difference across major versions. Depending on the surface area of the package you use they may or may not affect you. I'd love to be able to run tests against multiple major versions of our peer dependencies to make this more verifiable and not have to pick one! At the moment it relies on developers knowing/manually checking that the peer dep range is still valid.

mxdvl: I don’t know how to achieve it in a Node environement easily. I think the best would be to have a test matrix where you actuall install all of these lowest major versions in CI and run all the tests against your package. I can imagine this being a maintenance hassle.

arelra: > You should always pin your devDependencies to the lowest possible match

aiui this is because you don't know which version consumers of your repo might install?

Given that the only consumers of braze-components are DCR and Frontend can we upgrade all the ranges to the latest versions as we have insight and control of all the consumers?

ecguardian (approved):
👍

PR #388, opened 2023-02-07
Fix canRender file to include new Newsletter Epic templates

KaliedaRik:
## What does this change?
A small update to make sure the new Newsletter Epic templates created in #385 will be rendered to End Users when required.

### DIFF
--- a/src/canRender.ts
+++ b/src/canRender.ts
@@ -1,12 +1,31 @@
+// Banners
+// --------------------------------------------
 import {
     COMPONENT_NAME as APP_BANNER_NAME,
     canRender as appBannerCanRender,
 } from './AppBanner/canRender';
+
 import {
     COMPONENT_NAME as DIGITAL_SUBSCRIBER_APP_BANNER_NAME,
     canRender as digitialSubCanRender,
 } from './DigitalSubscriberAppBanner/canRender';
+
+import {
+    COMPONENT_NAME as BANNER_WITH_LINK_NAME,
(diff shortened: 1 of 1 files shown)


src/canRender.ts
```diff
@@ -1,12 +1,31 @@
...
+// Default Epics
+// --------------------------------------------
 import { COMPONENT_NAME as EPIC_NAME, canRender as epicCanRender } from './Epic/canRender';
+
+import {
+    COMPONENT_NAME as EPIC_WITH_IMAGE_NAME,
+    canRender as epicWithImageCanRender,
+} from './EpicWithSpecialHeader/canRender';
+
+// Old newsletter Epics
```
tjmw, line 27: Totally fine to organise things this way, but I'm wondering if we need to separate out the newsletter epics into old and new? I think it's just the name format which has changed between the two groups?

tjmw (approved):
This looks good!

Sorry I didn't spot that this was needed on the original PR to add the two new components. It's annoying that this isn't caught it an automated way currently (by the type system or by a test).

Perhaps adding a test would be a simple solution for now. For example if a component exists but isn't included in the mapping table in here then that causes a test failure? (Not for this PR, just thinking how to avoid this in future).

src/canRender.ts
KaliedaRik: In a few weeks we'll be moving to a more generic self-serve epic which should get rid of the need for Epic Newsletter naming strategies

PR #389, opened 2023-02-08
Link CODE storybook to CODE preview for previewing rendering

tjmw:
## What does this change?

In storybook there's a button in the header to preview the components in place on the preview site. The CODE instance of storybook links to prod preview but it feels more useful to link to code preview.

## How to test

Click the button in CODE storybook and the component will render in CODE preview.

### DIFF
--- a/.storybook/utils.ts
+++ b/.storybook/utils.ts
@@ -45,4 +45,15 @@ const getLoginUrl = () => {
     }
 };
 
-export { getGridUrl, getImageSigningUrl, getLoginUrl };
+const getPreviewUrl = () => {
+    switch (getEnvironment()) {
+        case 'LOCAL':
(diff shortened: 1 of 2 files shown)


.storybook/utils.ts
```diff
@@ -45,4 +45,15 @@ const getLoginUrl = () => {
     }
 };
 
-export { getGridUrl, getImageSigningUrl, getLoginUrl };
+const getPreviewUrl = () => {
+    switch (getEnvironment()) {
+        case 'LOCAL':
```
KaliedaRik, line 50: Small question: why do we call it `LOCAL` here, not `DEV`? Do we use the `getEnvironment()` function more widely (in which case: don't fix) or just here?

KaliedaRik (approved):
Left a tiny comment, but it's not a blocker. Approved.

.storybook/utils.ts
tjmw: I _think_ it's entirely internal to this file (`getEnvironment` isn't exported). Definitely happy to rename, but I think a separate PR makes sense.

PR #394, opened 2023-02-17
Newsletter banners part 1: refactor storybook stories

KaliedaRik:
## What does this change?
This PR refactors storybook stories for newsletter-related components. This preliminary work will the refactor of newsletter-related components so that they can be shared between banners and epics easier.

PR #396, opened 2023-02-17
Newsletter banners part 2: refactor banner common functionality

KaliedaRik:
## What does this change?
This PR refactors current banner components to place shared functionality into its own file. This preliminary work will make the creation of a new banner component (for newsletters) easier

Note that this PR builds on https://github.com/guardian/braze-components/pull/394


src/AppBanner/index.tsx
```diff
@@ -57,36 +52,24 @@ export const AppBanner = (props: Props): ReactElement | null => {
 
     const [showBanner, setShowBanner] = useState(true);
 
-    const onCloseClick = (
-        evt: React.MouseEvent<HTMLButtonElement, MouseEvent>,
-        internalButtonId: number,
-    ): void => {
-        evt.preventDefault();
+    const logToBrazeAndOphan = (internalButtonId: number): void => {
+        logBannerCloseToOphan(
```
tjmw, line 56: I wonder if this should be slightly renamed or reorganized? I think it's potentially bit confusing that the `logBannerCloseToOphan` function takes the `logButtonClickWithBraze` function as an argument (i.e. the scope is greater than the function name suggests, since it also reports to Braze). What do you reckon?

I think perhaps there's also an extra level of function application here that we don't need as `logToBrazeAndOphan` invokes `logBannerCloseToOphan` but doesn't do anything else. Perhaps the imported function could be renamed to something more general like `logClickToBrazeAndOphan` and invoked directly by `onCloseAction` etc?

src/bannerCommon/bannerActions.ts
```diff
@@ -0,0 +1,64 @@
...
+    }, deps);
+}
+
+export type OnCloseClick = (
+    evt: React.MouseEvent<HTMLButtonElement, MouseEvent>,
+    internalButtonId: number,
+) => void;
+
+// Function to log banner close events
+export const logBannerCloseToOphan = (
```
tjmw, line 41: I _think_ maybe this is more generic than closing the banner, in that it can be used to log _any_ banner button click (i.e. there's nothing specific to closing banners in here I don't think). What do you think of renaming to reflect that? Also relates to comment above about the name here making is appear that this is only related to Ophan when in fact it also logs clicks to Braze.

src/AppBanner/index.tsx
KaliedaRik: This is a bit embarrassing ... I refactored all this out (in Part 4, when adding the new newsletter banner) after I discovered we already have the code to handle it in the `utils/tracking.ts` file. I have no idea why banners weren't already using that code - insights on any edge case I'm missing would be very welcome!

src/bannerCommon/bannerActions.ts
KaliedaRik: See comment above re discovering code already exists in `utils/tracking.ts` file, and changing to use that code in PR Part 4

src/AppBanner/index.tsx
tjmw: Ahh interesting. From the commit which last touched `utils/tracking.ts`:

```
This includes a refactoring to expose a single trackClick callback to
the individual epic components, instead of submitComponentEvent and
logButtonClickWithBraze separately. This keeps the underlying code
simpler and also DRYs things up a bit. I've only done this refactoring
for epics here. Banners will be done in a separate PR.
```

I guess this bit never happened - "Banners will be done in a separate PR". So it'd be good to make banners and epics consistent. Let me know if I can help with that.

PR #397, opened 2023-02-17
Newsletter banners part 3: refactor newsletter common components

KaliedaRik:
## What does this change?
In order to enable single-click newsletter signup in the banner, we need to make use of existing functionality currently coded in Braze epics. This PR identifies and collects these components into a new file, and updates existing epic components to use them.

Note that this PR builds on https://github.com/guardian/braze-components/pull/396


src/newsletterCommon/sharedComponents.tsx
```diff
@@ -0,0 +1,236 @@
+import React, { useState } from 'react';
+import { css, ThemeProvider, keyframes } from '@emotion/react';
+import { Button, buttonThemeBrandAlt, SvgClock } from '@guardian/source-react-components';
+import type { TrackClick } from '../utils/tracking';
+
+import { news, neutral, body, textSans } from '@guardian/source-foundations';
+
+// Newsletter CTA button
+// -------------------------------------------------------
```
KaliedaRik, line 9: At some point I think it will make sense to separate this, alongside the other 2 components defined in this file, into their own file

src/BrazeEndOfArticleComponent.ts
```diff
@@ -1,5 +1,8 @@
 import React from 'react';
 
+// Common newsletter imports
+import { NewsletterSubscribeCallback } from './newsletterCommon/sharedComponents';
```
tjmw, line 4: `NewsletterSubscribeCallback` is a type, so I guess this should probably be a type import?

```suggestion
import type { NewsletterSubscribeCallback } from './newsletterCommon/sharedComponents';
```

src/newsletterCommon/sharedComponents.tsx
tjmw: Until then, what do you think about renaming this to `index.tsx`? Since this is the entirety of the common newsletter code right now. That would also give us nice imports like:

```typescript
import { CTA } from './newsletterCommon';

Comment (KaliedaRik) [reply src/newsletterCommon/sharedComponents.tsx:9]: Happy to do this

File: src/BrazeEndOfArticleComponent.ts (hunk 2 above)
Comment (KaliedaRik) [reply src/BrazeEndOfArticleComponent.ts:4]: We can make this happen

File: src/newsletterCommon/sharedComponents.tsx [hunk 3]
```diff
@@ -0,0 +1,236 @@
...
+const ctaStyles = {
```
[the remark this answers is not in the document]
KaliedaRik: This is a very interesting question! I think in my ideal world I would like to keep the component the same for both banner and epic, but adapt the component so that it can receive styling data from each and respond accordingly. Like this 
Screenshot 2023-03-01 at 10 52 44

src/newsletterCommon/sharedComponents.tsx
KaliedaRik: Done

src/newsletterCommon/sharedComponents.tsx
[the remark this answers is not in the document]
tjmw: Cool yeah that sounds like a good strategy. I think I tend to prefer a bit of duplication until that becomes an issue, but I don't feel super strongly so happy to go with your judgement!

PR #398, opened 2023-02-17
Newsletter banners part 4: create newsletter banner component

KaliedaRik:
## What does this change?
This PR creates, and plumbs in, the new Newsletter Banner
- Refactors banners to bring their tracking into alignment with epic tracking
- Adds new BannerNewsletter folder and files
- Sets up the new banner to use the `one-click subscribe` button, as used by epic newsletter components

Note that this PR builds on https://github.com/guardian/braze-components/pull/397 

### Screenshots
Screenshot 2023-03-02 at 13 43 46

Screenshot 2023-03-02 at 13 43 59


src/DigitalSubscriberAppBanner/index.test.tsx
```diff
@@ -22,19 +21,15 @@ describe('DigitalSubscriberAppBanner', () => {
             const { getByText } = render(
                 <DigitalSubscriberAppBanner
                     {...baseProps()}
-                    submitComponentEvent={logButtonClickWithOphan}
+                    trackClick={logButtonClickWithOphan}
```
tjmw, line 24: Nitpick: shall we rename `logButtonClickWithOphan` to reflect the broader use of this callback now (i.e. it doesn't just log to Ophan).

src/DigitalSubscriberAppBanner/index.tsx
```diff
@@ -1,39 +1,38 @@
...
+import type { TrackClick } from '../utils/tracking';
 import { COMPONENT_NAME } from './canRender';
 export { COMPONENT_NAME };
 
 export type BrazeMessageProps = {
+    ophanComponentId?: string;
     header?: string;
     body?: string;
+    cta?: string;
+    imageUrl?: string;
```
tjmw, line 12: I think it's confusing to add these as props here as they aren't supported by this component? (i.e. the image and cta are hard coded in this case).

src/BrazeBannerComponent.ts
```diff
@@ -1,32 +1,39 @@
...
+    DigitalSubscriberAppBanner,
+} from './DigitalSubscriberAppBanner';
+import { COMPONENT_NAME as APP_BANNER_NAME, AppBanner } from './AppBanner';
+import { COMPONENT_NAME as BANNER_WITH_LINK_NAME, BannerWithLink } from './BannerWithLink';
+import { COMPONENT_NAME as BANNER_NEWSLETTER_NAME, BannerNewsletter } from './BannerNewsletter';
 
 export type CommonBannerComponentProps = {
     componentName: string;
+    brazeMessageProps: BrazeMessageProps;
+    subscribeToNewsletter: NewsletterSubscribeCallback;
```
tjmw, line 27: Just a reminder (although TypeScript will let you know too!) that this will need to be added as a prop provided by the platform on DCR and frontend, like we do for newsletters [here](https://github.com/guardian/dotcom-rendering/blob/main/dotcom-rendering/src/web/components/SlotBodyEnd/BrazeEpic.tsx#L130).

src/BannerWithLink/index.tsx
```diff
@@ -22,25 +17,22 @@ export type BrazeMessageProps = {
 };
 
 export type Props = {
-    logButtonClickWithBraze: BrazeClickHandler;
-    submitComponentEvent: (componentEvent: OphanComponentEvent) => void;
-    ophanComponentId?: string;
```
tjmw, line 27: I think we need to add this back here to reflect that the `ophanComponentId` passed from Braze in the message extras is used by the component.

src/AppBanner/index.tsx
```diff
@@ -32,18 +28,14 @@ export type BrazeMessageProps = {
 import type { ButtonTheme } from '@guardian/source-react-components';
 
 export type Props = {
-    logButtonClickWithBraze: BrazeClickHandler;
-    submitComponentEvent: (componentEvent: OphanComponentEvent) => void;
-    ophanComponentId?: string;
```
tjmw, line 37: As with `BannerWithLink` I don't think we want to lose the `ophanComponentId` arg here. I _think_ this will also remove the need to cast with `as string` below.

src/DigitalSubscriberAppBanner/index.tsx
KaliedaRik: It raises the question: why do we have/support the `DigitalSubscriberAppBanner` template? The only difference between this and `AppBanner` is that the CTA and URL are hardcoded which, moving to a greater self-serve world, is something we do not want to encourage?

src/DigitalSubscriberAppBanner/index.test.tsx
KaliedaRik: Renamed to `logButtonClick` across relevant banner test files which mock such a function

src/DigitalSubscriberAppBanner/index.tsx
KaliedaRik: In the Accessibility PR, which builds on this PR, I've factored out the local `BrazeMessageProps` type from each banner `index.tsx` to an all-encompassing imported type (maybe it should be an `interface`?) - which you can see defined here: https://github.com/guardian/braze-components/pull/400/files#diff-3341aff8c9584e4e301a155c9d5b5f58b589524f09dd49b5743eef8515798994 ... but I'm happy to take out unnecessary items from the local declarations in this PR.

src/BannerWithLink/index.tsx
KaliedaRik: Sorry, I don't understand this comment. The functions have been replaced by the Epic's `trackClick` function - which we agreed was something that had been planned a while back but never implemented? As a consequence the `ophanComponentId` attribute moves into the `brazeMessageProps` object

src/DigitalSubscriberAppBanner/index.tsx
KaliedaRik: Done

tjmw: > In the Accessibility PR, which builds on this PR, I've factored out the local BrazeMessageProps type from each banner index.tsx to an all-encompassing imported type (maybe it should be an interface?)

Hmm wonder if that's what we want. I'll use epics here as an example but the same is true for banners. High level, there is a common interface which all Braze epics have which is defined here: https://github.com/guardian/braze-components/blob/main/src/BrazeEndOfArticleComponent.ts#LL49-L56C3 So that's everything the platform will provide to the high level `BrazeEndOfArticleComponent` component. At this level the `brazeMessageProps` are very generic and represent the data we get from Braze. Then downstream of that each individual component specifies narrower types, which outline the props that the specific component wants. So taking the `BrazeDownToEarthNewsletterEpic` as an example, it's props are defined here: https://github.com/guardian/braze-components/blob/main/src/DownToEarthNewsletterEpic/index.tsx#L19, so this narrows down and makes more specific the higher level props. Mixing all possible props that each component might want in a higher level type/interface at the top feels like it might be confusing as it'll be harder to know what props/values from Braze an individual component cares about. What do you think? (Happy to pair on any of this if it's useful or if I'm not making sense!)

tjmw: Hey yes I think moving to more generic/configurable components on the whole is a good strategy. The approach we've used in the past on this project is create more bespoke/specific templates and then if they prove useful to create something more generic (which is how `DigitalSubscriberAppBanner` -> `AppBanner` happened). Great question about `DigitalSubscriberAppBanner`, I'd be surprised if that is still in use anywhere. If we can get our marketing colleagues to confirm it's unused we could remove it (I think that probably belongs in its own PR).

src/BrazeBannerComponent.ts
tjmw: I think I'd do that work in a single PR on each of the web platforms. So once the newsletter banner changes are in a release, create a PR on each of frontend and DCR which 1) bumps the version of braze-components and 2) starts threading through the `subscribeToNewsletter` prop for banners. I can help with integration testing etc in CODE when we're ready to do that!

src/BannerWithLink/index.tsx
tjmw: Hey sorry I didn't explain this very well. Having dug a bit more it looks like for this component the `ophanComponentId` would have always come from Braze and so would be in the `brazeMessageProps` just as you said. So removing seems like a good clean up 👍

You probably saw that for the two older banner components before we'd figured out how tracking would work we were falling back on the actual React component name for the `ophanComponentId` (just to give some context on why that's happening).

src/BannerWithLink/index.tsx
```diff
@@ -22,25 +17,22 @@ export type BrazeMessageProps = {
...
     brazeMessageProps: BrazeMessageProps;
+    trackClick: TrackClick;
 };
 
 const BannerWithLink: React.FC<Props> = (props: Props) => {
     const {
-        logButtonClickWithBraze,
-        submitComponentEvent,
         brazeMessageProps: {
+            ophanComponentId = COMPONENT_NAME,
```
tjmw, line 27: For `BannerWithLink` the `ophanComponentId` is something that we require marketing to provide. It's marked as required in Storybook and the `canRender` enforces this. So it's probably not needed to provide the default/fallback here.
```suggestion
            ophanComponentId,
```

src/BannerNewsletter/index.tsx
```diff
@@ -0,0 +1,114 @@
...
+export type Props = {
+    brazeMessageProps: BrazeMessageProps;
+    trackClick: TrackClick;
+    subscribeToNewsletter: NewsletterSubscribeCallback;
+};
+
+const BannerNewsletter: React.FC<Props> = (props: Props) => {
+    const {
+        brazeMessageProps: {
+            ophanComponentId = COMPONENT_NAME,
```
tjmw, line 29: Similar to `BannerWithLink`, I don't think we need to provide a fallback/default here as we enforce this as a required param in the `canRender` (and mark it as required for marketing in Storybook).

```suggestion
            ophanComponentId,
```

src/BannerWithLink/index.tsx
KaliedaRik: Will remove

src/BannerNewsletter/index.tsx
KaliedaRik: Will remove

src/BannerWithLink/index.tsx
KaliedaRik: ... Done

src/BannerNewsletter/index.tsx
KaliedaRik: ... Done

PR #399, opened 2023-02-21
Fix the build

tjmw:
## What does this change?

Fix the build. We started seeing errors in the build on the build-storybook step which look like they're related to the Node version according to Google:

`Error: error:0308010C:digital envelope routines::unsupported`

It turns out we weren't honouring the .nvmrc since migrating to GitHub actions for the build. Updating the workflow to correctly use the .nvmrc seems to get things working.

## Accessibility


KaliedaRik (approved):
(FingersCrossed) that this solves the build issue

PR #400, opened 2023-02-22
Braze Banner accessibility fixes

KaliedaRik:
## What does this change?
Hackday project to make Braze banners properly accessible

Refactors AppBanner by:
- Pulling constituent parts of the AppBanner out into their own components
- Cleaned up the styling
- Added accessibility functionality

Note that this PR builds on https://github.com/guardian/braze-components/pull/398


src/styles/bannerCommon.ts
```diff
@@ -51,6 +51,20 @@ export const styles = {
         }
     `,
 
+    topBar: css`
```
KaliedaRik, line 54: This file can be deleted once the other two banners are moved over to the new components

PR #404, opened 2023-03-06
Remove old digisubs banner code

KaliedaRik:
## What does this change?
This PR removes the Digital Subscriptions banner from the code base. This is an old banner which has been superseded by the App and With-link banners. Marketing have confirmed that they no longer use it.


tomrf1 (approved):
great!

PR #408, opened 2023-04-03
React 18

tomrf1:
A department-wide upgrade is underway.
DCR + Frontend haven't yet been upgraded.
This PR upgrades to react 18 in braze-components.

I have tested components:
1. in storybook
2. locally with DCR `main` branch, using `yarn link`

It's a strange upgrade because DCR actually includes preact client-side ([aliased to react](https://github.com/guardian/dotcom-rendering/blob/main/dotcom-rendering/scripts/webpack/webpack.config.js#L40))

By specifying the peerDependency as `17.0.2 || 18.2.0` we widen the range of supported versions.

### DIFF
--- a/package.json
+++ b/package.json
@@ -59,11 +59,11 @@
     "@storybook/components": "^6.5.4",
     "@storybook/preset-typescript": "^3.0.0",
     "@storybook/react": "^6.5.4",
-    "@testing-library/jest-dom": "^5.11.4",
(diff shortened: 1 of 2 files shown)


tjmw (approved):
I think this looks good 👍  As discussed it'd be good to test this against versions of DCR & frontend in CODE after the upgrade (though as you mention, I'm not sure what the difference will be given that Preact is shipped to the client).

PR #409, opened 2023-04-04
Bump @guardian/source-* libs and TypeScript

arelra:
## What does this change?

Bump:
typescript@~4.9.5 (as per csnx version)
@guardian/source-react-components-development-kitchen@^10.0.1
@guardian/source-react-components@^12.0.0
@guardian/source-foundations@^10.0.0

Also:
Fix TS errors

This is to align peer deps for: https://github.com/guardian/frontend/pull/25892 

This should be a released as a major version increment as peer deps have been bumped.

### DIFF
--- a/package.json
+++ b/package.json
@@ -96,9 +96,9 @@
   "peerDependencies": {
     "@emotion/react": "^11.1.2",
     "@guardian/libs": "^14.0.0",
-    "@guardian/source-foundations": "^9.0.0",
-    "@guardian/source-react-components": "^11.2.0",
-    "@guardian/source-react-components-development-kitchen": "^9.2.0",
+    "@guardian/source-foundations": "^10.0.0",
+    "@guardian/source-react-components": "^12.0.0",
+    "@guardian/source-react-components-development-kitchen": "^10.0.1",
     "react": "17.0.2"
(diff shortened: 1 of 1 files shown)


tjmw (approved):
This looks good to me! ✅ 

Is there an equivalent DCR PR to guardian/frontend#25892 which will perform the same source library version bumps we see here? I think if possible it'd be great to merge & release this braze-components change as close as possible to when the frontend/DCR branches are merged. Otherwise we end up with a change in main which will then be included in any other future releases off of main that we might want to do before the platform PRs are ready (I suppose this is something we could manage better with changesets).

If it's useful there's also the option to create a pre-release/beta version off of this branch.

arelra:
> Is there an equivalent DCR PR to [guardian/frontend#25892](https://github.com/guardian/frontend/pull/25892) which will perform the same source library version bumps we see here? 

Thanks @tjmw . I think DCR is already at these versions?

https://github.com/guardian/dotcom-rendering/blob/main/dotcom-rendering/package.json#L77-L79

Once braze-components is released I'll bump in both DCR & Frontend asap hopefully that should minimise the window between any further updates to main?

PR #410, opened 2023-04-13
Update authored epic component to make it more configurable

KaliedaRik:
## What does this change?
Marketing have requested that we update the `EpicWithSpecialHeader` component to allow them to self-serve the editing of the component's image and byline copy and the accompanying header copy.

## Images

**Storybook: before change**
Screenshot 2023-04-13 at 13 04 31

**Storybook: after change**
Screenshot 2023-04-13 at 13 04 48

## Accessibility
This PR makes the `` attribute configurable - previously it was served as a zero-length string

### DIFF
--- a/src/Epic/index.tsx
+++ b/src/Epic/index.tsx
@@ -6,6 +6,7 @@ import { COMPONENT_NAME, canRender, parseParagraphs } from './canRender';
 export { COMPONENT_NAME };
 import { replaceNonArticleCountPlaceholders } from './placeholders';
 import { TrackClick } from '../utils/tracking';
+import { HeaderSection } from '../EpicWithSpecialHeader';
 
 // Custom styles for <a> tags in the Epic content
 const linkStyles = css`
@@ -72,13 +73,18 @@ export type BrazeMessageProps = {
     remindMeConfirmationHeaderText?: string;
     remindMeConfirmationText?: string;
     hidePaymentIcons?: string;
+    authoredEpicHeader?: string;
+    authoredEpicImageUrl?: string;
+    authoredEpicImageAltText?: string;
+    authoredEpicBylineName?: string;
+    authoredEpicBylineCopy1?: string;
+    authoredEpicBylineCopy2?: string;
 };
 
 export type EpicProps = {
     ophanComponentId?: string;
     brazeMessageProps: BrazeMessageProps;
     countryCode?: string;
-    headerSection?: React.ReactNode;
     trackClick: TrackClick;
 };
 
@@ -94,9 +100,14 @@ export const Epic: React.FC<EpicProps> = (props: EpicProps) => {
             remindMeConfirmationText,
             ophanComponentId,
             hidePaymentIcons,
+            authoredEpicHeader,
+            authoredEpicImageUrl,
+            authoredEpicImageAltText,
+            authoredEpicBylineName,
+            authoredEpicBylineCopy1,
+            authoredEpicBylineCopy2,
         },
         countryCode,
-        headerSection,
         trackClick,
     } = props;
 
@@ -114,7 +125,16 @@ export const Epic: React.FC<EpicProps> = (props: EpicProps) => {
         <ThemeProvider theme={brand}>
             <div css={styles.epicWrapper}>
                 <section css={styles.epicContainer}>
-                    {headerSection}
+                    {authoredEpicImageUrl && (
+                        <HeaderSection
+                            authoredEpicImageUrl={authoredEpicImageUrl}
+                            authoredEpicImageAltText={authoredEpicImageAltText}
+                            authoredEpicHeader={authoredEpicHeader}
+                            authoredEpicBylineName={authoredEpicBylineName}
+                            authoredEpicBylineCopy1={authoredEpicBylineCopy1}
+                            authoredEpicBylineCopy2={authoredEpicBylineCopy2}
+                        />
+                    )}
 
                     <span css={styles.heading}>{heading}</span>
                     {paragraphs.map((text, index) => (
--- a/src/EpicWithSpecialHeader/index.stories.tsx
+++ b/src/EpicWithSpecialHeader/index.stories.tsx
@@ -8,6 +8,7 @@ import {
     buildEpicParagraphDocs,
 } from '../storybookCommon/argTypes';
 import { BrazeMessageProps } from '../Epic';
+import { grid, withGrid } from '../../.storybook/grid/withGrid';
 
 const NUMBER_OF_PARAGRAPHS = 9;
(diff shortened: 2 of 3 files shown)


tomrf1 (approved):
great!

src/EpicWithSpecialHeader/index.tsx
```diff
@@ -52,34 +48,62 @@ const headerStyles = {
     `,
 };
 
+type HeaderSectionProps = {
+    authoredEpicImageUrl?: string;
```
tomrf1, line 52: should all these fields be mandatory in this component?

src/Epic/index.tsx
```diff
@@ -72,13 +73,18 @@ export type BrazeMessageProps = {
     remindMeConfirmationHeaderText?: string;
     remindMeConfirmationText?: string;
     hidePaymentIcons?: string;
+    authoredEpicHeader?: string;
+    authoredEpicImageUrl?: string;
+    authoredEpicImageAltText?: string;
+    authoredEpicBylineName?: string;
+    authoredEpicBylineCopy1?: string;
+    authoredEpicBylineCopy2?: string;
```
tjmw, line 81: Since these are now props which can be passed through to the regular `Epic` component, I don't think there's now any need for the `EpicWithSpecialHeader` component. I think this is highlighted by the definition of the `EpicWithSpecialHeader` component being simply a call to the `Epic` component:

```js
export const EpicWithSpecialHeader: React.FC<EpicProps> = (props: EpicProps) => <Epic {...props} />;
```

src/EpicWithSpecialHeader/index.tsx
tjmw: I think probably for this component some of these should be required (i.e. if it'd look odd to render the header with no byline name, then we should make it required). I think TypeScript will force us to do this, but in the `Epic` component, we should only render `HeaderSection` if the required props are available.

src/EpicWithSpecialHeader/index.stories.tsx
```diff
@@ -8,20 +8,58 @@ import {
...
 import { BrazeMessageProps } from '../Epic';
+import { grid, withGrid } from '../../.storybook/grid/withGrid';
 
 const NUMBER_OF_PARAGRAPHS = 9;
 const paragraphDocs = buildEpicParagraphDocs(NUMBER_OF_PARAGRAPHS);
 
 export default {
     component: 'EpicWithSpecialHeader',
     title: 'EndOfArticle/EpicWithSpecialHeader',
+    decorators: [withGrid],
```
tjmw, line 19: The grid image picker is a custom storybook plugin we wrote to help marketing interact with the Grid and build a signed `i.guim.co.uk` image URL. I think there are possibly elements of it which are hardcoded around an image size suitable for a banner which may not be appropriate here (i.e. generating images which are too large for the small header image size used by this component).

src/EpicWithSpecialHeader/index.tsx
tomrf1: Is it possible to nest the mandatory fields under an optional field in the main epic props type? E.g.
```
type BrazeMessageProps = {
  buttonText?: string;
  ...
  authorHeader?: {
    authoredEpicImageUrl: string;
    authoredEpicImageAltText: string;
    authoredEpicHeader: string;
    authoredEpicBylineName: string;
    authoredEpicBylineCopy1: string;
    authoredEpicBylineCopy2: string;
  }
```

Then `HeaderSection` doesn't have to deal with optional fields.

tjmw: Ah yep that's nice 👍 (but remember that these will all be coming through as a flat structure of key/value pairs from Braze, so we'll have to do the nesting here)

src/EpicWithSpecialHeader/index.stories.tsx
KaliedaRik: I'm not too concerned about image size as it seems there's functionality in the grid plugin to create a new crop of the image if size is an issue. I think this can be solved by making sure people using the tool know about the option to create new crops

Screenshot 2023-04-17 at 12 23 23

The tool also gives a warning about over-small images, which can lead to poor quality image rendering in the browser/app

Screenshot 2023-04-17 at 12 24 19

I did see an issue with using an image of inappropriate (non-square) images - they can appear behind the close button. Adding an `autofocus` attribute to the close button could help to make it more prominent?

Screenshot 2023-04-17 at 12 17 39

Screenshot 2023-04-17 at 12 18 18

At the end of the day, the question is: are we willing to let Marketing do these checks for themselves when skinning the component with data in Storybook? Or do we want to limit their choices in various ways by hardcoding them into the component's `canRender` checks?

src/EpicWithSpecialHeader/index.tsx
KaliedaRik: I think I'd prefer the props to be flat? Hierarchical data (in my view) adds complexity where it's not essential and may trip up those who have to maintain components like this going forward?

src/Epic/index.tsx
KaliedaRik: For now I'd like to leave it as its own separate component as that's what Marketing expect it to be. There's also the issue where we want some of these new props to be mandatory for the byline epic, but can be ignored for the default epic and I don't know how to solve that in either storybook or in the default epic's `canRender` check

src/EpicWithSpecialHeader/index.tsx
tomrf1: Sure. My original point was that if fields are necessary to render a component, it's simpler + more correct for that component to receive them as non-optional. But since braze doesn't support nesting there's probably not much to be gained by nesting here

KaliedaRik: Updated to check the imageUrl, altText and bylineName fields have been completed.

tjmw: > Updated to check the imageUrl, altText and bylineName fields have been completed.

Great 👍  I think the types here should probably reflect that, what do you think? I.e. right now the `HeaderSectionProps` type says that `authoredEpicImageUrl` is optional, but in the JSX we use it as though it's non-optional (using it as the source for a `img` tag for example.

KaliedaRik: Done

tomrf1: I'm confused about why some of these props are optional and some not - it looks like they're all being rendered?

KaliedaRik: Pushed an update - does that address your concern?

tomrf1 (approved):
👍 looks good!

PR #414, opened 2023-06-15
Legitimate image url check for epics

KaliedaRik:
## What does this change?
When we updated the epicWithSpecialImage header we accidentally left out the check which determines whether the supplied image url comes from a permitted Guardian source. This PR rectifies the situation

## How to test
Test in Storybook


tomrf1 (approved):
👍

PR #416, opened 2023-06-28
New self-serve styleable banner

KaliedaRik:
## What does this change?
Currently the styling of Braze banners is hardcoded
Marketing colleagues have requested that we create a new banner more amenable to self-serve styling, in particular the ability to:
+ Set the banner background colour
+ Set the banner text colour
+ Set the highlighted text background colour
+ Set the highlighted text colour
+ Set the vertical position of any image (`center`, `bottom`)
+ Set the primary CTA background colour
+ Set the primary CTA text colour

Additional styling-related changes have also been requested but, for this MVP proof-of-concept banner, they are out of scope for this PR.

This PR will also make no effort to improve the accessibility of any banner - though that work will need to be addressed sooner rather than later.

To complete the work, we will also need to create related PRs in the following repos
+ **Frontend** - to bump the BC version dependency
+ **DCR** - to recreate the new banner in Storybook there, and bump the BC version dependency

## Screenshot examples
[TODO]

## Accessibility
See above comment


src/styles/bannerCommon.ts
```diff
@@ -238,3 +238,261 @@ export const styles = {
...
+export const selfServeStyles = (style: StyleData) => {
+    return {
+        wrapper: css`
+            box-sizing: border-box;
+            display: flex;
+            flex-direction: column;
+            justify-content: center;
+            align-items: center;
+            width: 100%;
+            background-color: ${style.styleBackground};
```
tjmw, line 263: I wonder if we should be sanitising this user input before interpolating into the styles? I'm not worried about malicious intent but more accidentally pasting bad content in which could break the page. What do you think? It's possible to add further styles here by adding `;` to the prop/value passed from Braze, for example changing `styleBackground` to `red; transform: rotate(90deg)` gives: 
Screenshot 2023-06-30 at 15 03 44

Perhaps we could add some simple validation to the `canRender` method to ensure only values which match `rbd(\d?, \d?, \d?)` or similar? We could even parse the supplied string into something with the following type, if validation passes:

```
interface ColourStyle {
  r: number,
  g: number,
  b: number
}
```

This would allow us to give richer type information in the function signature of `selfServeStyles` (`ColourStyle` tells the developer more than `string`) and would make it impossible to pass arbitrary strings in to be interpolated into the styles.

src/styles/bannerCommon.ts
```diff
@@ -238,3 +238,261 @@ export const styles = {
         }
     `,
 };
+
+interface StyleData {
```
tjmw, line 242: I don't _think_ these are used anywhere other than the new banner component currently, so I'd vote to colocate them with that code, rather than putting them in bannerCommon. What do you think?

tjmw (commented):
Nice work, this is really exciting! I've left a few comments, see what you think (happy to pair or chat about them).

A final overall question - since `StyleableBannerWithLink` now does everything (and more) than `BannerWithLink`, could we simply update `BannerWithLink` to call `StyleableBannerWithLink`? I guess with a view to eventually deprecate it.

src/StyleableBannerWithLink/index.stories.tsx
```diff
@@ -0,0 +1,180 @@
...
+            disable: false,
+        },
+    },
+    argTypes: {
+        ...coreArgTypes,
+        ...ophanComponentIdArgType,
+        styleBackground: {
+            name: 'styleBackground',
+            type: { name: 'string', required: false },
+            description: 'Banner background color - defaults to "rgb(237, 237, 237)"',
```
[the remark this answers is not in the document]
KaliedaRik: I'll check and correct if this is a bug

src/styles/bannerCommon.ts
KaliedaRik: My thinking is that as soon as we deploy this, Marketing will come back and ask us for a similarly configurable newsletter banner. Both the existing NewsletterBanner and BannerWithLink use the exported `style` object on this page. A new StyleableNewsletterBanner could make use of the `selfServeStyles` function ... so my gut instinct is to leave the function here. I know it's future-proofing for something that may not happen - but does that make sense?

src/styles/bannerCommon.ts
tjmw: That's true, there are many different formats, but I don't think we need to support them here. I think it would be a worthwhile trade-off to limit the flexibility to one (or two) formats because of the extra safety it would give and validation it would allow us to do.

KaliedaRik:
`could we simply update BannerWithLink to call StyleableBannerWithLink?` - yes! I've coded the new attributes to have defaults that match the `BannerWithLink` styling (except for the image position which is by default vertically centred but for the current image needs to be `bottom`

The more hardcoded banners and epics we can delete, the easier the (eventual) migration of these components to DCR will be!

src/styles/bannerCommon.ts
KaliedaRik: A fix to prevent malicious/accidental css has been pushed up - also fixes the default CSS colors issue.

KaliedaRik: To fix the issue of malformed color strings I've pushed up an additional fix which works by applying default colors in the CSS immediately before the required color (which may not conform to CSS color definitions). If the required color is a malformed color string then the CSS engine will skip it and the displayed colour will fallback to the default color

src/StyleableBannerWithLink/index.tsx
```diff
@@ -0,0 +1,188 @@
...
+            imageAltText = '',
+            imagePosition = 'center',
+            styleClose,
+            styleCloseBackground,
+            styleCloseHover,
+        },
+        trackClick,
+    } = props;
+
+    const styles = selfServeStyles(
```
tjmw, line 78: I wonder if we can simplify the interface to `selfServeStyles` a bit so that we don't need to specify defaults twice? Once in the inline call to `cssInjectionCheck` and then again as the second argument. What if `selfServeStyles` called `cssInjectionCheck` itself, and then use the second `defaultColors` arg to know what to default to? That way all of the "defaulting" would be happening in one place, rather than being spread out? What do you think?

src/styles/bannerCommon.ts
```diff
@@ -2,239 +2,291 @@ import { css } from '@emotion/react';
...
-        ${from.desktop} {
-            max-width: 980px;
-            flex-direction: row;
+            display: flex;
+            flex-direction: column;
+            justify-content: center;
+            align-items: center;
+            width: 100%;
+            background-color: ${defs.styleBackground};
+            background-color: ${style.styleBackground};
```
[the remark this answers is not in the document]
KaliedaRik: Done

src/StyleableBannerWithLink/index.tsx
KaliedaRik: TS error fixed

src/AppBanner/index.tsx
```diff
@@ -33,21 +48,19 @@ export type Props = {
...
     } = props;
 
-    if (!canRender(props.brazeMessageProps)) {
-        return null;
-    }
+    const commonStyles = selfServeStyles(props.brazeMessageProps, defaultColors);
 
     const [showBanner, setShowBanner] = useState(true);
 
-    if (!canRender(props.brazeMessageProps)) {
```
tomrf1, line 47: why were there two `canRender` checks before? Mistake?

src/styles/bannerCommon.ts
```diff
@@ -1,240 +1,336 @@
...
+    'styleButtonBackground',
+    'styleButtonHover',
+    'styleClose',
+    'styleCloseBackground',
+    'styleCloseHover',
+];
+
+type Styles = keyof StyleData;
+
+export const selfServeStyles = (userVals: Extras, defs: StyleData) => {
```
tomrf1, line 37: is `defs` defaults? And if so can we call it that please

src/styles/bannerCommon.ts
```diff
@@ -1,240 +1,336 @@
...
+    canvas.height = 1;
+    const ctx = canvas.getContext('2d', { willReadFrequently: true });
+
+    defKeys.forEach((key) => {
+        const userVal = userVals[key];
+        const defVal = defs[key];
+
+        // If user val is undefined, or an empty string, use default val
+        if (userVal == null || !userVal) {
+            style[key] = defVal;
```
tomrf1, line 52: is this line redundant, because we initialised `style` as `Object.assign({}, defs)`?

src/styles/bannerCommon.ts
```diff
@@ -1,240 +1,336 @@
...
+                    flag = false;
+                }
+            }
+            // Check for legitimate CSS color string values
+            if (flag && ctx && colorStringStyles.includes(key)) {
+                ctx.fillStyle = 'transparent';
+                ctx.fillStyle = item;
+                ctx.clearRect(0, 0, 1, 1);
+                ctx.fillRect(0, 0, 1, 1);
+                const iData = ctx.getImageData(0, 0, 1, 1);
```
tomrf1, line 78: would `getImageData` fail if the colour was invalid?

KaliedaRik: `getImageData` will always return an imageData object. If the colour is invalid then the returned pixel will be transparent black `[0, 0, 0, 0]` which is how we can identify the user supplied colour string is malformed

src/styles/bannerCommon.ts
KaliedaRik: Okay - will fix

src/AppBanner/index.tsx
KaliedaRik: Yep - fixed. I've also moved the check to the top of the function in each banner

src/styles/bannerCommon.ts
tomrf1: ok, please can you add this in a comment so it's clear to people like me

src/styles/bannerCommon.ts
KaliedaRik: Done

src/styles/bannerCommon.ts
KaliedaRik: Done

src/styles/bannerCommon.ts
```diff
@@ -1,240 +1,313 @@
...
+        // If user val is undefined, or an empty string, use default val
+        if (userVal != null && userVal.length > 0) {
+            let flag = true;
+
+            // Protect against CSS injection
+            const item = userVal.split(';')[0].trim();
+
+            // Check for legitimate CSS color string values
+            // - we only support `#abcdef` color format
+            if (flag && colorStringStyles.includes(key)) {
```
tomrf1, line 54: I'm struggling with the meaning of `flag` - why are we checking it here? What does it mean?

KaliedaRik: Function refactored

tjmw: Oh! I'm still seeing `flag` here, is there an un-pushed commit maybe?

tomrf1 (approved):
great!

src/styles/bannerCommon.ts
```diff
@@ -1,240 +1,310 @@
...
+export const selfServeStyles = (userVals: Extras, defaults: StyleData) => {
+    const style: StyleData = Object.assign({}, defaults);
+    const defKeys: Styles[] = Object.keys(defaults) as Styles[];
+    const regex = new RegExp(/^#([A-Fa-f0-9]{6})$/);
+
+    defKeys.forEach((key) => {
+        const userVal = userVals[key];
+
+        // If user val is undefined, or an empty string, use default val
+        if (userVal != null && userVal.length > 0) {
```
tjmw, line 46: Would a simple `if (userVal)` be simpler/more readable here? I think that would cover the undefined and empty string cases.

src/styles/bannerCommon.ts
```diff
@@ -1,240 +1,310 @@
...
+            // Protect against CSS injection
+            const item = userVal.split(';')[0].trim();
+
+            // Protect against null or empty user strings
+            if (item == null || !item.length) {
+                return;
```
tjmw, line 52: Super minor, but I wonder, since we're using an early `return` here, should we be consistent and do that elsewhere in this function instead of nested `if`s? I think I find it a bit more readable when one of those two approaches is used rather than mixing them.

tjmw (approved):
I left a couple of minor questions/suggestions, but this looks good!

src/styles/bannerCommon.ts
KaliedaRik: if statements un-nested, as suggested

src/styles/bannerCommon.ts
KaliedaRik: No longer relevant

PR #418, opened 2023-07-21
Add card payment symbols option to styleable banner CTA button

KaliedaRik:
## What does this change?
A small update to the new styleable banner which will allow Marketing colleagues to add the payment icons to the main CTA button

## Screenshots
Screenshot 2023-07-21 at 16 37 07

Screenshot 2023-07-21 at 16 37 26

### DIFF
--- a/src/styles/bannerCommon.ts
+++ b/src/styles/bannerCommon.ts
@@ -204,6 +204,17 @@ export const selfServeStyles = (userVals: Extras, defaults: StyleData) => {
             margin-right: ${space[3]}px;
         `,
 
+        primaryButtonWrapper: css`
+            margin: ${space[4]}px ${space[2]}px ${space[1]}px 0;
+            display: flex;
+            flex-wrap: wrap;
+            align-items: center;
+
+            &.hidden {
+                display: none;
+            }
+        `,
+
         primaryButton: css`
             margin-right: ${space[3]}px;
             color: ${style.styleButton};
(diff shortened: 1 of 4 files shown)


tomrf1 (approved):
👍 looks good.
One day it would be nice to parse these bools-as-strings at the entry point so we don't have to do string comparisons all over the place!

src/styles/bannerCommon.ts
```diff
@@ -204,6 +204,17 @@ export const selfServeStyles = (userVals: Extras, defaults: StyleData) => {
             margin-right: ${space[3]}px;
         `,
 
+        primaryButtonWrapper: css`
+            margin: ${space[4]}px ${space[2]}px ${space[1]}px 0;
+            display: flex;
+            flex-wrap: wrap;
+            align-items: center;
+
+            &.hidden {
```
tjmw, line 213: Out of interest, when does this hidden class get added?

KaliedaRik: You know ... I don't think it ever does! I copied this CSS from `src/Epic/ContributionEpicButtons` and, looking at that file there's nothing to indicate a need for a `hidden` class because the payment icons don't even render unless the boolean prop is set. I think we can remove this in a future tidy-up PR?

tjmw: Ah cool yeah that's what I was wondering, like you said if the icons are hidden the payment icons component isn't even rendered. Wasn't sure if I was missing something though! Future tidy up PR sounds good!

KaliedaRik: Another change to consider at some point is the image we use for the payment icons (because: not sure the image has a transparent background which may make them appear weird in a banner with a non-grey background). RRCP banners and epics are a mess when it comes to the icons - some use an image, others use some hardcoded SVG. Sadly there doesn't appear to be a definitive version of the icons in the style guide, and the icons in Source don't match ours.

PR #421, opened 2023-08-01
Refactor components in preparation for reminders work

KaliedaRik:
## What does this change?
This small PR reorganises some components that are about to be shared between Braze Epics and Banners (principally the contributions-with-reminders buttons) into a new `components` folder

## Testing
Storybook should be sufficient

## Release
I suggest we don't publish the repo after this PR is merged into main?


tjmw (approved):
Looks good! And nice to see Chromatic report no changes as expected.

> I suggest we don't publish the repo after this PR is merged into main?

Sounds good 👍

PR #422, opened 2023-08-02
Refactor Epic reminder button to use RRCP equivalent functionality

KaliedaRik:
## What does this change?
Marketing colleagues have asked us to refactor Braze banner and epic components so that they use the reminder functionality currently available on web messaging components rather than the canvas-based 30-day functionality we use with Braze.

The ask requires us to know what the user's email address is, which is not immediately available to Braze components. We achieve this by threading a `fetchEmail` function from DCR into the Braze component which we can then invoke when the user clicks on the reminder CTA.

Work to thread the `fetchEmail` function into Braze components was completed in this PR: https://github.com/guardian/braze-components/pull/423. This PR builds the new Reminder CTA button and updates the default Epic component to use it.

Related DCR PR: https://github.com/guardian/dotcom-rendering/pull/8561

## Navigating this PR
**Thread a new DCR-supplied attribute `isCodeEnvironment` into Banner/Epic components**
- src/BrazeBannerComponent.ts
- src/BrazeEndOfArticleComponent.ts

**The new reminders functionality**
- src/logic/reminders.ts
- src/utils/env.ts

**Separate the Contribution and Reminder CTA buttons into their own components**
- src/components/PrimaryCtaButton.tsx
- src/components/ContributionCtaButton.tsx
- src/components/ReminderCtaButton.tsx

**Separate the loading dots animation into its own component file**
- src/components/CtaLoadingDotsAnimation.tsx
- src/newsletterCommon/index.tsx

**Refactor the Epic component to use new reminder functionality**
- src/Epic/index.tsx
- src/Epic/index.test.tsx
- src/Epic/index.story.tsx
- src/Epic/failure.stories.tsx

All the remaining file changes are minor fixes to Storybook files

## How to test
- Create a canvas in Braze-dev which will trigger the Braze epic to be added to an article
- Merge this PR into main, publish the repo to NPM, and update the DCR PR to use the latest version
- Deploy the DCR PR to CODE
- Navigate to the Guardian CODE site, view articles until the epic appears
- Click on the Reminder CTA and confirm it works
- Check that a reminder has been created for the user in the appropriate system(s)

## Images

Before user clicks:
Screenshot 2023-10-31 at 11 03 46

When user clicks, the flashing dots animation replaces the button while the fetchEmail and setReminder functions are in flight (not shown)

... If reminder is successfully set:
Screenshot 2023-10-31 at 11 04 00

... If an error is encountered:
Screenshot 2023-10-31 at 11 04 14


src/logic/reminders.ts
```diff
@@ -0,0 +1,82 @@
+import { isTest } from '../utils/env';
+
+const CREATE_ONE_OFF_REMINDER_ENDPOINT = 'https://support.theguardian.com/reminders/create/one-off';
```
tjmw, line 3: Should this vary based on the environment? I guess there's a CODE equivalent? For other similar API endpoints, we thread these through from the platform (which has the stage awareness), for example the subscribe to newsletter endpoint.

src/components/ReminderCtaConfirmation.tsx
```diff
@@ -66,17 +95,8 @@ export const ReminderCtaConfirmation = ({
...
-            <p css={successTextStyles}>
-                {remindMeConfirmationText} You can manage your email preferences in the My Account
-                area,{' '}
-                <a href="https://manage.theguardian.com/email-prefs">
-                    emails and marketing section
-                </a>
-                .
-            </p>
+            <h4 css={successHeadingStyles}>Something went wrong</h4>
+            <p css={successTextStyles}>Sorry, we were not able to sign you up for a reminder.</p>
```
tjmw, line 99: Maybe there's a different name we could use for `successTextStyles` since these definitely aren't success messages!

src/components/ContributionButtonWithReminder.tsx
```diff
@@ -76,14 +83,20 @@ const PrimaryButton = ({ buttonUrl, buttonText, trackClick }: PrimaryButtonProps
 );
 
 interface RemindMeButtonProps {
+    disabled: boolean;
```
tjmw, line 86: I'm a bit confused about when this button will be disabled? I can see that the following is passed:

```
reminderStatus !== ReminderStatus.Editing
```

What does `ReminderStatus.Editing` mean here? (i.e. what state does it represent?) Would it be fair to think of `Editing` as the default state when a user hasn't taken any action (yet)?

src/components/ContributionButtonWithReminder.tsx
```diff
@@ -93,60 +106,60 @@ const RemindMeButton = ({ remindMeButtonText, onClick }: RemindMeButtonProps) =>
...
     buttonUrl: string;
-    remindMeButtonText?: string;
-    remindMeConfirmationText?: string;
-    remindMeConfirmationHeaderText?: string;
+    reminderStage?: ReminderStage;
+    reminderOption?: string;
     trackClick: (buttonId: number) => void;
     hidePaymentIcons?: string;
 }
-type SectionState = 'DEFAULT' | 'REMINDER_CONFIRMED' | 'REMINDER_CONFIRMATION_CLOSED';
```
tjmw, line 102: I really liked having this as a single authority on the different states you can be in regarding reminders in this component. Now I think it's spread across two states: `reminderStatus` (which has type `ReminderStatus`) and also `reminderConfirmationOpen` (which is a boolean). I wonder if it's be worth combining them? Could `reminderConfirmationOpen` be represented as a new member of the `ReminderStatus` enum perhaps?

It feels like spreading across two states opens the possibility of representing impossible situations. For example what would it mean for `reminderConfirmationOpen` to be `true`, but `reminderStatus` to be `Editing`? It feels like there are combinations which should ever happen, so [it'd be nice to make them impossible to represent](https://www.youtube.com/watch?v=IcgmSRJHu_8).

src/components/ContributionButtonWithReminder.tsx
```diff
@@ -93,60 +106,60 @@ const RemindMeButton = ({ remindMeButtonText, onClick }: RemindMeButtonProps) =>
...
+                    remindMeButtonText={reminderCta}
                     onClick={() => {
                         trackClick(REMIND_ME_BUTTON_INTERNAL_ID);
-                        setSectionState('REMINDER_CONFIRMED');
+
+                        setReminderStatus(ReminderStatus.Submitting);
+
+                        createReminder({
+                            reminderPeriod,
+                            email: 'TODO',
```
tjmw, line 154: Previously we didn't have to worry about knowing the user's email. Because they were logged in, we knew which Braze user they were, and the reminder email was triggered directly in Braze.

But now we need to pass the email to the reminders API. I think for signed in users it's available on the platform, I believe for SDC users it pre-populates the email input so perhaps we could use the same data? It looks like a `fetchEmail` prop is passed from the platform: https://github.com/guardian/dotcom-rendering/blob/f7fc75d7f7b9bb446fc2131ec46feef18a0f50e8/dotcom-rendering/src/components/StickyBottomBanner/ReaderRevenueBanner.tsx#L386 (this is for a banner, but same thing I think).

src/logic/reminders.ts
KaliedaRik: Agreed! Will look into it

src/components/ReminderCtaConfirmation.tsx
KaliedaRik: There's a lot of names we can change - particularly for css-related code. This will become particularly important for when we try to thread this functionality into the styleable banner code. Will investigate.

src/components/ContributionButtonWithReminder.tsx
KaliedaRik: See comment below - I don't think we need this functionality, which is used in RRCP Banner/Epic for signed-out users who have to manually enter their email address.

src/components/ContributionButtonWithReminder.tsx
KaliedaRik: The reminder CTA functionality in RRCP Banner/Epic is messy and (I think) overcomplicated for our needs. This is because in RRCP the button has to cope with both signed-in users where the code fetches the user's email address using the `fetchEmail()` function supplied by DCR, and signed-out users where clicking the button opens a microform where users can manually enter their email which then needs to be sanitised etc.

We only need to handle signed-in users, so should be able to use `fetchEmail()` once we get DCR to pipe it into our Epic (and Banner).

I really like the way the Braze Epic/Banner newsletter signup CTA button works (with the animated dots) - if we use a similar approach here, then when the user clicks the Remind Me Later button they will see the animated dots, then a success/fail message will display in place of all the CTA buttons. I'll try to build something for us to look at in a new PR

src/logic/reminders.ts
KaliedaRik: A fix is in place - we now get the environment as an attribute from DCR

src/components/ReminderCtaConfirmation.tsx
KaliedaRik: File has been removed as we no longer need it

src/components/ContributionButtonWithReminder.tsx
KaliedaRik: We no longer use this journey flow, thus suggested fix no longer required

src/components/ContributionButtonWithReminder.tsx
KaliedaRik: The user journey has changed significantly - see screenshots above

src/components/ReminderCtaButton.tsx
```diff
@@ -0,0 +1,186 @@
...
+        </ThemeProvider>
+    </div>
+);
+
+interface SmallPrintProps {
+    month: string;
+}
+
+const SmallPrint = ({ month }: SmallPrintProps) => (
+    <div css={styles.smallPrint}>
```
tjmw, line 102: Has there been design input on this bit? In the screenshots the font doesn't look very Guardian-y!

src/components/ReminderCtaButton.tsx
```diff
@@ -0,0 +1,186 @@
...
+        setRemindState('IN_PROGRESS');
+
+        fetchEmail()
+            .then((email) => {
+                if (email) {
+                    return createReminder({
+                        reminderPeriod,
+                        email,
+                        reminderPlatform: 'WEB',
+                        reminderComponent: 'EPIC',
```
tjmw, line 141: I wonder if it'd be useful to be able to distinguish between the RRCP epic and Braze epic?

src/logic/reminders.ts
```diff
@@ -0,0 +1,82 @@
...
+
+    return {
+        reminderCta: `Remind me in ${monthName}`,
+        reminderPeriod: `${year}-${paddedMonth}-01`,
+        reminderLabel: `${monthName} ${year}`,
+    };
+};
+
+export const createReminder = (signupData: OneOffSignupRequest): Promise<void> => {
+    const url = REMINDER_ENDPOINT_PROD;
```
tjmw, line 60: Could we use `REMINDER_ENDPOINT_PROD` directly rather than assign to a const?

src/logic/reminders.ts
```diff
@@ -0,0 +1,82 @@
...
+};
+
+export interface ReminderFields {
+    reminderCta: string;
+    reminderLabel: string;
+    reminderPeriod: string;
+    reminderOption?: string;
+}
+
+export const buildReminderFields = (today: Date = new Date()): ReminderFields => {
```
tjmw, line 44: Given that the logic here is slightly tricky, is it worth adding some jest tests to act as documentation (and give us confidence if we ever want to make changes to this code)?

src/Epic/index.tsx
```diff
@@ -151,20 +164,26 @@ export const Epic: React.FC<EpicProps> = (props: EpicProps) => {
...
-                                ophanComponentId: ophanComponentId as string,
-                            })
-                        }
-                    ></ContributionButtonWithReminder>
+                    <div css={styles.buttonWrapperStyles}>
+                        <ContributionCtaButton
+                            buttonText={buttonText as string}
+                            buttonUrl={buttonUrl as string}
+                            hidePaymentIcons={hidePaymentIcons as string}
+                            internalButtonId={PRIMARY_BUTTON_INTERNAL_ID}
```
tjmw, line 172: Would it be better to have these internal tracking IDs be part of the relevant button component (rather than pass them in)?

src/logic/reminders.ts
```diff
@@ -0,0 +1,82 @@
+import { isTest } from '../utils/env';
+
+const REMINDER_ENDPOINT_PROD = 'https://support.theguardian.com/reminders/create/one-off';
+// const REMINDER_ENDPOINT_CODE = 'https://support.code.dev-theguardian.com/reminders/create/one-off';
+
+export type ReminderStatus = 'DEFAULT' | 'IN_PROGRESS' | 'SUCCESS' | 'FAILURE';
```
tjmw, line 6: Given that this is a UI state concern, I wonder if it'd be better to define this inline in the component? I don't _think_ anything else would need to access this type. What do you think?

src/BannerWithLink/index.stories.tsx
```diff
@@ -77,7 +77,9 @@ const StoryTemplate = (args: BrazeMessageProps & { componentName: string }): Rea
                 logButtonClickWithBraze={(internalButtonId) => {
                     console.log(`Button with internal ID ${internalButtonId} clicked`);
                 }}
+                subscribeToNewsletter={() => mockSubscribe('')}
```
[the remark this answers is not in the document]
KaliedaRik: I think because DCR is passing the newsletter subscribe function through to banner component as well as epic, since we added the newsletter banner? But why it was not picked up before ... I don't know!

src/Epic/index.tsx
KaliedaRik: We have a ticket in Trello to sort out Braze banner/epic message tracking - maybe look at this as part of that work?

src/components/ReminderCtaButton.tsx
KaliedaRik: No design input. I pasted a screenshot into the TX chat a while back o get feedback. Will look at making sure the correct font is being used here

src/components/ReminderCtaButton.tsx
KaliedaRik: Suggest it to TX tomorrow? But maybe not for this PR?

src/logic/reminders.ts
```diff
@@ -0,0 +1,82 @@
+import { isTest } from '../utils/env';
+
+const REMINDER_ENDPOINT_PROD = 'https://support.theguardian.com/reminders/create/one-off';
+// const REMINDER_ENDPOINT_CODE = 'https://support.code.dev-theguardian.com/reminders/create/one-off';
```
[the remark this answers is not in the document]
KaliedaRik: Yes ... will fix

src/logic/reminders.ts
KaliedaRik: It's used by both the reminder button, and the newsletter subscribe button (from which I stole it)

src/logic/reminders.ts
KaliedaRik: I can have a go ...

src/logic/reminders.ts
KaliedaRik: Yes - this is a result of an attempt earlier in the history of this PR to differentiate endpoints based on code environment - but as RRCP only uses PROD for reminders I was persuaded to abandon that effort. Will update.

src/Epic/index.tsx
tjmw: I think that ticket is more a concern about sending the right tracking data. Here I'm talking more about internal code structure/organisation. If we're always passing in a static hardcoded value maybe it'd reduce complexity to have that value be part of the component (rather than adding a prop to parameterise it)?

src/logic/reminders.ts
tjmw: Oh are you sure it's being used by the newsletter subscribe button? I can't see that import in this PR. I don't feel super strongly, but I'd say on balance it's probably not worth the slight additional complexity to share this between the two types of buttons (newsletter signup and reminders) and would opt for duplicating. If we do want to share as you say renaming would be good (and maybe defining it in a file outside of this reminders specific code?)

src/logic/reminders.ts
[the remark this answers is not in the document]
KaliedaRik: Fixed

src/logic/reminders.ts
KaliedaRik: Fixed

src/logic/reminders.ts
KaliedaRik: Ah ... I pushed a renaming fix, and moved to a types file, before reading this comment.

src/Epic/index.tsx
KaliedaRik: Updated as suggested

src/BrazeBannerComponent.ts
```diff
@@ -26,6 +26,7 @@ export type CommonBannerComponentProps = {
     brazeMessageProps: BrazeMessageProps;
     subscribeToNewsletter: NewsletterSubscribeCallback;
     fetchEmail: FetchEmail;
+    isCodeEnvironment: boolean;
```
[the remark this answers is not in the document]
KaliedaRik: Fixed

src/logic/reminders.ts
KaliedaRik: Done - let me know if we need to test other things here

src/components/ReminderCtaButton.tsx
KaliedaRik: Fonts fixed. Images updated in the PR description

src/UKNewsletterEpic/failure.stories.tsx
```diff
@@ -74,6 +74,7 @@ const StoryTemplate = (
                     );
                 }}
                 fetchEmail={() => mockFetchEmail()}
+                s
```
tjmw, line 77: I think this might be a typo?

src/Epic/index.tsx
```diff
@@ -151,20 +161,24 @@ export const Epic: React.FC<EpicProps> = (props: EpicProps) => {
...
-                    ></ContributionButtonWithReminder>
+                    <div css={styles.buttonWrapperStyles}>
+                        <ContributionCtaButton
+                            buttonText={buttonText as string}
+                            buttonUrl={buttonUrl as string}
+                            hidePaymentIcons={hidePaymentIcons as string}
+                            ophanComponentId={ophanComponentId as string}
+                            trackClick={trackClick}
+                        />
+                        {reminderStage && reminderOption && (
```
tjmw, line 172: Sorry I didn't notice this before - is it right that a `reminderOption` should always be specified when scheduling a reminder? I think in SDC, `reminderOption` isn't usually specified.

src/UKNewsletterEpic/failure.stories.tsx
KaliedaRik: That's just weird - removed!

src/Epic/index.tsx
KaliedaRik: Good spot. Removed!

tjmw (approved):

PR #423, opened 2023-08-10
Add fetchEmail callback to Braze components

KaliedaRik:
## What does this change?
RRCP reminder CTA buttons make use of a fetchEmail function supplied by DCR to locate the email address of a signed-in user. This PR replicates that functionality for Braze components - though no component makes use of it (will appear in the upcoming Reminders CTA work for Epics and Banners)

Related DCR PR: [TODO]

## Testing
TODO: write a test to see if we can retrieve an email of a given user? DCR functionality seems to rely on a cookie being present in the browser whose value it can feed into its `lazyFetchEmailWithTimeout` function - not sure how to replicate that locally in BC repo for testing


tjmw (commented):
This looks good! 👍 

I'm slightly surprised that there are no TypeScript errors reported in the build since a new required prop has been added to `BrazeEndOfArticleComponent` and `BrazeBannerComponent` and we're not passing it through in the stories.

Weirdly it _is_ reported in VSCode for me in all of the story files. Here's an example:

Screenshot 2023-08-14 at 10 08 32

src/utils/StorybookWrapper.tsx
```diff
@@ -27,6 +27,16 @@ export const mockSubscribe: NewsletterSubscribeCallback = (newsletterId) => {
     });
 };
 
+export const mockRemind: FetchEmail = () => {
```
tjmw, line 30: Very minor (feel free to ignore) - what do you think about a slightly different name here, say `mockFetchEmail`? Just to align more closely with what the mock is actually doing (it's not doing any reminding!)

tjmw (approved):

PR #424, opened 2023-09-18
Styleable banner hotfixes

KaliedaRik:
## What does this change?
Marketing reported 2 issues with the new styleable banner:
1 - the image does not center
2 - the highlighted text background color expands to fill the block rather than just highlight the text

We need to fix these issues so they can release a banner as part of the new Europe edition launch

## Screenshots

**Before fix**
Screenshot 2023-09-18 at 15 57 06

**After fix**
Screenshot 2023-09-18 at 15 57 38

### DIFF
--- a/src/StyleableBannerWithLink/index.tsx
+++ b/src/StyleableBannerWithLink/index.tsx
@@ -105,7 +105,7 @@ const StyleableBannerWithLink: React.FC<Props> = (props: Props) => {
 
                         {highlight ? (
                             <>
-                                <br />
+                                <br />&nbsp;<br />
                                 <strong css={styles.highlight}>{highlight}</strong>
                             </>
                         ) : null}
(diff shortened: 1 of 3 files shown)


src/StyleableBannerWithLink/index.tsx
```diff
@@ -105,6 +105,8 @@ const StyleableBannerWithLink: React.FC<Props> = (props: Props) => {
 
                         {highlight ? (
                             <>
+                                <br />
```
tjmw, line 108: I wonder if it would be better to use a container with margin to manage the spacing here? Does this give the right spacing at all screen widths? (Just thinking - if not, I think it's easier to adapt to different screen widths with a marging vs a `br` tag).

tjmw (approved):

PR #425, opened 2023-09-19
Fix spacing on BannerWithLink and AppBanner

tjmw:
## What does this change?

This updates the markup in BannerWithLink and AppBanner to match StyleableBannerWithLink so that the shared styling is applied in the same way. Previously after #424 was merged we lost a bit of spacing between the body text and highlighted text on other banners as the container didn't exist in the markup.

## How to test

Storybook!

## Accessibility


KaliedaRik (approved):
I've never liked the `br` tag between body and highlight text. They should always have been either their own `p` tags or, like RRCP, just a space so all fits in one paragraph

PR #428, opened 2023-10-25
Support Node 20.9 LTS as an engine

mxdvl:
## What does this change?

Add support for Node 20 (Iron) as an engine. This is also an LTS version of Node [as of today](https://nodejs.org/en/blog/release/v20.9.0).

## How to test

Everything should run as-is, but [you can confirm using this online checker](https://jubianchi.github.io/semver-check/#/^18.15%20||%20^20.9/20.9).

Consumer of this package should not be constrained to stay on Node 18, which is now in maintenance mode.

It would probably make more sense to use an unbound upper range, [such as `>=18.15`](https://jubianchi.github.io/semver-check/#/%3E%3D18.15/20.9).

## Images

N/A


tjmw (approved):
Looks good, sorry for the delay!

@mxdvl I'm not certain what the relationship should be between engines/node and our `.nvmrc`. Would it make sense to bump `.nvmrc` to `20.9` to reflect the fact we're saying that this package can be consumed by projects running on `^20.9`? Out of interest I pulled your branch and ran the tests/storybook on `20.9` and everything seemed to work fine.

I guess we can do a quick and to end test on CODE by deploying guardian/dotcom-rendering#9308 ?

mxdvl:
> @mxdvl I'm not certain what the relationship should be between engines/node and our `.nvmrc`

I believe the engine field should be [treated similarly to that of a peer dependency](https://github.com/guardian/recommendations/blob/main/npm-packages.md#pin-the-lowest-possible-version-of-peerdependencies-in-your-packages-devdependencies), which would mean you would have your `.nvmrc` pointing to the lowest possible version of your range. In this case, that would be `18.15.0`. I don’t think we ought to be too prescriptive about this.

How do I get this to create a `PATCH` release?

tjmw:
> I believe the engine field should be [treated similarly to that of a peer dependency](https://github.com/guardian/recommendations/blob/main/npm-packages.md#pin-the-lowest-possible-version-of-peerdependencies-in-your-packages-devdependencies), which would mean you would have your .nvmrc pointing to the lowest possible version of your range. In this case, that would be 18.15.0. I don’t think we ought to be too prescriptive about this.

Ah good to know, that makes sense. Thanks for explaining 👍 

> How do I get this to create a PATCH release?

Sadly we're still cutting releases/publishing to NPM for this project manually off of main (changesets coming soon!) @KaliedaRik just merged a change to `main` which hasn't been published yet, but I think will be a major bump. @KaliedaRik maybe you could work with Max to get the releases out? I'd suggest: create release 16.0.0 off of `main` as it is now (before this PR is merged) and then `16.0.1` with Max's Node change. What do you think?

mxdvl:
> I'd suggest: create release 16.0.0 off of `main` as it is now (before this PR is merged) and then `16.0.1` with Max's Node change. What do you think?

I’m not too fussy, it’s only been about a week since Node 20 has reached LTS. I’ll merge this now and simply await for the next package bump in `dotcom-rendering`.

PR #430, opened 2023-10-27
Refactor to make the reminder button configurable

KaliedaRik:
## What does this change?
This PR refactors some code which will make the work of adding the reminders button to the styleable banner component easier.

The changes are:
+ A new `styles/colorData.ts` file to colocate colour styling related types and functions in one place.
+ The colour-related types and objects, and functionality to validate user-supplied colour strings, has been moved out of `styles/bannerCommon.ts` into the new file. The `AppsBanner`, `BannerNewsletter`, `BannerWithLink`, `Epic` and `StyleableBannerWithLink` components have been updated to reflect these changes, as has the `ContributionCtaButton` component.
+ The `components/CtaLoadingDotsAnimation` component has been updated to return a function to output the SVG, which can accept a colour string argument. The `newsletterCommon/index.tsx` and `components/ReminderCtaButton.tsx` components have been updated to use the function
+ Most of the refactor in this PR has been to the `components/ReminderCtaButton.tsx` component. These changes mean that the component can now be styled by user-defined colour strings, with default values for those strings based on the existing colours of the reminder button. The colour types and objects now in the `styles/colorData.ts` file have been updated to include these new optional attributes.

## How to test
All chromatic tests should pass straight away as this PR is an internal refactor which shouldn't affect graphical output.


src/components/CtaLoadingDotsAnimation.tsx
```diff
@@ -36,7 +36,7 @@ const loadingAnimCss = css`
     }
 `;
 
-export const LoadingDots = (): React.ReactElement => {
+export const buildLoadingDots = (fillStyle = '#707070'): React.ReactElement => {
```
tjmw, line 39: Would it be more conventional maybe to have this continue to be named `LoadingDots` (it is a component after all, not a function which returns a component)? The `fillStyle` could be passed as an (optional) prop, which would allow it to be called in the normal way: `<LoadingDots fillStyle="#121212" />`. What do you think?

src/components/ReminderCtaButton.tsx
```diff
@@ -1,123 +1,106 @@
...
-        :hover {
-            background-color: ${neutral[86]} !important;
-        }
-    `,
-    smallPrint: css`
-        color: ${neutral[0]};
-        margin-top: ${space[2]}px;
-        ${body.small()};
-    `,
+const getButtonStyles = (userVals: Extras, defaults: ColorStylesData) => {
```
tjmw, line 25: I wonder if we could be more descriptive/specific with the types of these args. `Extras` is very generic and doesn't provide much information about what might be available. My understanding is `getColors` takes some colours which _might_ be provided and some defaults which _will_ be provided and comes up with a new instance of `ColorStylesData` which is guaranteed to have specific fields set (if we can't get the colour from the user provided values then fall back on the defaults).

What if the type of the user provided styles was called something like `UserDefinedColorStylesData` and it looked something like this (clipped for brevity):

```
export interface UserDefinedColorStylesData {
    styleBackground?: string;
    styleHeader?: string;
    styleBody?: string;
}
```

So any of those fields may be there, but could also be missing. I think this tells us more about the possible user provided styles than `Extras` does. Then we could define `ColorStylesData` as the type which is guaranteed to have all these fields:

```
type ColorStylesData = Required<UserDefinedColourStylesData>;
```

Then the type of `getColors` becomes: `(userVals: UserDefinedColorStylesData, defaults: ColorStylesData): ColorStylesData`.

src/components/ReminderCtaButton.tsx
```diff
@@ -1,123 +1,106 @@
...
+    const styles = getColors(userVals as Extras, defaults);
```
tjmw, line 26: Currently the return value of `getColours` is `ColorStylesData` and all of the fields in that type are optional - but we interpolate them into the CSS styles below as though they are guaranteed to be present. I think returning a type where the fields are guaranteed to be there would be a win in terms of type safety (as in the comment above). What do you reckon? Happy to chat this over/pair if it's useful!

src/components/ReminderCtaButton.tsx
```diff
@@ -1,123 +1,106 @@
...
-const SmallPrint = ({ month }: SmallPrintProps) => (
-    <div css={styles.smallPrint}>
-        We will send you a maximum of two emails in {month}. To find out what personal data we
-        collect and how we use it, view our{' '}
-        <a href="https://manage.theguardian.com/email-prefs">Privacy Policy</a>.
-    </div>
-);
-
 interface ReminderCtaButtonProps {
+    reminderComponent: ReminderComponent;
```
tjmw, line 82: 👏 Nice!

tjmw (commented):
Nice refactoring! Good shout to make this a separate PR. I've left some comments/suggestions on the types, see what you think.

src/components/CtaLoadingDotsAnimation.tsx
KaliedaRik: Current code matches the factory approach we have to generating CSS, but I'm happy to turn this into a component. Will action.

src/components/ReminderCtaButton.tsx
KaliedaRik: Will take this suggestion forward. We can pair if I run into TS issues?

If we can get rid of `Extras` ... I wouldn't cry. I think there's another type with the same definition somewhere.

src/components/ReminderCtaButton.tsx
KaliedaRik: As mentioned above, will give it a go

src/components/ReminderCtaButton.tsx
tjmw: > We can pair if I run into TS issues?

Definitely! 👍

src/components/CtaLoadingDotsAnimation.tsx
KaliedaRik: Done!

src/components/ReminderCtaButton.tsx
KaliedaRik: I've given it a go. Discuss tomorrow?

src/components/ReminderCtaButton.tsx
KaliedaRik: See above

src/components/ReminderCtaButton.tsx
KaliedaRik: Thx

src/AppBanner/index.tsx
```diff
@@ -10,14 +10,15 @@ import type { TrackClick } from '../utils/tracking';
 
 import { AppStore } from '../assets/app-store';
 import { PlayStore } from '../assets/play-store';
-import { StyleData, selfServeStyles } from '../styles/bannerCommon';
+import { BannerColorStyles } from '../styles/colorData';
+import { selfServeStyles } from '../styles/bannerCommon';
```
tjmw, line 14: I think a thing that's been confusing me about the configurable colours is that `AppBanner`, `BannerNewsletter` & friends, which don't support configuring the colours, still call `selfServeStyles`. I think that's because `selfServeStyles` serves two purposes:

1. Takes user provided styles, and integrates these with defaults (if, for example, the user provides invalid values, or if certain colours are missing)
2. Generates CSS from a the colour styles object.

In the case of the banners which don't have configurable colours, we only need the second of those two purposes. This could be done in a follow up PR, but I think it would improve the understandability of the code to de-couple those two things. So instead of `selfServeStyles` taking user provided colours and defaults it just takes colours which are guaranteed to be present and returns the CSS.  It wouldn't have to care about where the colours came from (for `AppBanner` etc they'd be hardcoded, for `StyleableBannnerWithLink` they come from combining user provided colours with defaults). At the top level of `StyleableBannerWithLink` it'd have to do the combining of user provided colours and defaults, using `getColors`, and `selfServeStyles` wouldn't call `getColors`.

Thoughts?

src/styles/colorData.ts
```diff
@@ -0,0 +1,103 @@
...
+    'styleReminderButtonHover',
+    'styleReminderAnimation',
+    'styleClose',
+    'styleCloseBackground',
+    'styleCloseHover',
+];
+
+type ColorStylesType = keyof AllAvailableColorStyles;
+
+export function getColors(userVals: Extras, defaults: Partial<AllAvailableColorStyles>) {
```
tjmw, line 74: I wonder if using generics here would be a better way to describe the interface of `getColours`? I don't think `getColors` would ever be called with an object which had properties of both `BannerColorStyles` and `EpicColorStyles`, hence the need for `Partial`. If it was generic it called be called concretely with the specific relevant type acknowledging the fact that it would be called with a specific concrete type:

```
function getColors(userVals: Extras, defaults: T): T {
  // ...
}

// then called like:
getColors<BannerColorStyles>(userVals, defaultBannerColours)
getColors<EpicColorStyles>(userVals, defaultEpicColors)
/// etc
```

src/Epic/index.tsx
```diff
@@ -171,11 +179,13 @@ export const Epic: React.FC<EpicProps> = (props: EpicProps) => {
                         />
                         {reminderStage && (
                             <ReminderCtaButton
+                                reminderComponent="EPIC"
                                 reminderStage={reminderStage}
                                 reminderOption={reminderOption}
                                 ophanComponentId={ophanComponentId as string}
                                 trackClick={trackClick}
                                 fetchEmail={fetchEmail}
+                                userStyles={defaultColors}
```
tjmw, line 188: What about calling this prop `colors` or similar? I think `userStyles` isn't always accurate (in this case they're hardcoded for example).

src/components/ReminderCtaButton.tsx
```diff
@@ -1,123 +1,110 @@
...
-    smallPrint: css`
-        color: ${neutral[0]};
-        margin-top: ${space[2]}px;
-        ${body.small()};
-    `,
+const getButtonStyles = (
+    userVals: Partial<ReminderButtonColorStyles>,
+    defaults: ReminderButtonColorStyles,
+) => {
+    const styles = getColors(userVals as Partial<ReminderButtonColorStyles>, defaults);
```
tjmw, line 30: As in the comment above on `selfServeStyles` I wonder if it'd be better to decouple the two concerns here (merging user provided vals and defaults) and returning CSS. In some cases there won't be user provided styles as they're hardcoded/not configurable. So perhaps the places that need to can do the combining and then call this function with a complete set of colours.

tjmw (approved):
This is looking good. I've left a bunch of feedback, see what you think. If we wanted to action the points it could be done on this PR or in a follow up. Happy to chat if anything doesn't make sense (or if I've missed the point with any of it).

src/AppBanner/index.tsx
KaliedaRik: My long-term utopia is to have a single epic component, and a single banner component (both styleable, though in the case of the epic limited to eg CTA button colours) which Retention can setup/preview in Storybook. Though the competing vision of being able to use RRCP banner/epic designs from Braze is an even better utopia destination.

src/styles/colorData.ts
KaliedaRik: We've agreed to look at this in a future PR

src/Epic/index.tsx
KaliedaRik: I'll take this forward in the next PR

PR #433, opened 2023-11-15
Add reminder button to styleable banner

KaliedaRik:
## What does this change?
Adds the reminder button (as used in the Epic) to the styleable banner template

## How to test
Test in Storybook, and in CODE

This PR should only change the chromatic test for the Styleable banner

## Images
TODO

### DIFF
--- a/src/StyleableBannerWithLink/index.tsx
+++ b/src/StyleableBannerWithLink/index.tsx
@@ -1,11 +1,15 @@
 import React, { useState } from 'react';
-import { Button, LinkButton, SvgCross } from '@guardian/source-react-components';
+import { Button, SvgCross } from '@guardian/source-react-components';
 import { useEscapeShortcut, OnCloseClick, CLOSE_BUTTON_ID } from '../bannerCommon/bannerActions';
+import { ContributionCtaButton } from '../components/ContributionCtaButton';
+import { ReminderCtaButton } from '../components/ReminderCtaButton';
+import { ReminderStage } from '../logic/reminders';
 import type { TrackClick } from '../utils/tracking';
+import { FetchEmail } from '../types/dcrTypes';
 import { StyleableBannerColorStyles } from '../styles/colorData';
 import { selfServeStyles } from '../styles/bannerCommon';
 import { canRender, COMPONENT_NAME } from './canRender';
-import { PaymentIcons } from '../components/PaymentIcons';
+import { ColorValueHex } from '../logic/types';
 export { COMPONENT_NAME };
 
 export type BrazeMessageProps = {
@@ -16,6 +20,8 @@ export type BrazeMessageProps = {
     buttonText?: string;
     buttonUrl?: string;
     showPaymentIcons?: string;
+    reminderStage?: ReminderStage;
+    reminderOption?: string;
     includeReminderCta?: string;
     imageUrl?: string;
     imageAltText?: string;
@@ -58,6 +64,7 @@ const defaultColors: StyleableBannerColorStyles = {
 export type Props = {
     brazeMessageProps: BrazeMessageProps;
     trackClick: TrackClick;
+    fetchEmail: FetchEmail;
 };
 
 const StyleableBannerWithLink: React.FC<Props> = (props: Props) => {
@@ -74,14 +81,32 @@ const StyleableBannerWithLink: React.FC<Props> = (props: Props) => {
             buttonText,
             buttonUrl,
             showPaymentIcons = 'false',
+            reminderStage,
+            reminderOption,
             imageUrl,
             imageAltText = '',
             imagePosition = 'center',
         },
         trackClick,
+        fetchEmail,
     } = props;
 
-    const styles = selfServeStyles(props.brazeMessageProps, defaultColors);
+    const brazeProps = props.brazeMessageProps;
+
+    const styles = selfServeStyles(brazeProps, defaultColors);
+
+    const reminderStyles = {
+        styleReminderButton: brazeProps?.styleReminderButton as ColorValueHex,
+        styleReminderButtonBackground: brazeProps?.styleReminderButtonBackground as ColorValueHex,
+        styleReminderButtonHover: brazeProps?.styleReminderButtonHover as ColorValueHex,
+        styleReminderAnimation: brazeProps?.styleReminderAnimation as ColorValueHex,
+    };
+
+    const contributeStyles = {
+        styleButton: brazeProps?.styleButton as ColorValueHex,
+        styleButtonBackground: brazeProps?.styleButtonBackground as ColorValueHex,
+        styleButtonHover: brazeProps?.styleButtonHover as ColorValueHex,
+    };
 
     const [showBanner, setShowBanner] = useState(true);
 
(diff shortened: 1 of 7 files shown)


src/StyleableBannerWithLink/index.tsx
```diff
@@ -74,14 +81,32 @@ const StyleableBannerWithLink: React.FC<Props> = (props: Props) => {
...
-    const styles = selfServeStyles(props.brazeMessageProps, defaultColors);
+    const brazeProps = props.brazeMessageProps;
+
+    const styles = selfServeStyles(brazeProps, defaultColors);
+
+    const reminderStyles = {
+        styleReminderButton: brazeProps?.styleReminderButton as ColorValueHex,
+        styleReminderButtonBackground: brazeProps?.styleReminderButtonBackground as ColorValueHex,
+        styleReminderButtonHover: brazeProps?.styleReminderButtonHover as ColorValueHex,
+        styleReminderAnimation: brazeProps?.styleReminderAnimation as ColorValueHex,
```
tjmw, line 102: I'm confused about the casting happening here. These are all optional according to the `BrazeMessageProps` type above, but here we're casting to `ColorValueHex` without defaulting. That feels like it might cause problems downstream where the values maybe undefined at runtime but TypeScript won't know to warn us about this because we've said here they won't be (`ColorValueHex` is defined as a string template).

src/StyleableBannerWithLink/index.tsx
```diff
@@ -117,20 +142,28 @@ const StyleableBannerWithLink: React.FC<Props> = (props: Props) => {
...
-                                trackClick({
-                                    internalButtonId: 0,
-                                    ophanComponentId: ophanComponentId as string,
-                                })
-                            }
-                        >
-                            {buttonText}
-                        </LinkButton>
-                        {showPaymentIcons === 'true' && <PaymentIcons />}
+                        <ContributionCtaButton
```
tjmw, line 147: I wonder if it's confusing to use a component `ContributionCtaButton` here, since this banner is more generic than this (it could be a link to anything marketing want to promote)?

src/components/ContributionCtaButton.tsx
```diff
@@ -5,50 +5,76 @@ import { LinkButton } from '@guardian/source-react-components';
...
+            }
+        `,
+    };
+};
 
 interface ContributionCtaButtonProps {
     buttonText: string;
     buttonUrl: string;
-    hidePaymentIcons: string;
+    showPaymentIcons: boolean;
```
tjmw, line 48: Much nicer! 👍

src/components/ReminderCtaButton.tsx
```diff
@@ -134,28 +134,20 @@ export const ReminderCtaButton = ({
...
-                        disabled={false}
-                        ctaText={reminderCta}
-                    />
-                    <div css={styles.smallPrint}>
-                        We will send you a maximum of two emails in {reminderLabel}. To find out
-                        what personal data we collect and how we use it, view our{' '}
-                        <a href="https://manage.theguardian.com/email-prefs">Privacy Policy</a>.
-                    </div>
+                    <RemindMeButton buttonStyles={styles} onClick={onClick} ctaText={reminderCta} />
+                    {!isBanner && (
```
tjmw, line 138: Oh - do we not want to show this text on the banner? Is it a data privacy requirement to show it?

If it's okay to hide it on the banner, I'd be in favour of controlling it more explicitly i.e. with a prop called `hidePrivacyText` or similar. That feels more intuitive and robust to me than coupling to possible consumers in here. What do you think?

src/StyleableBannerWithLink/index.tsx
KaliedaRik: Shall we change the component's name to something like `PrimaryCtaButton`?

tjmw: That sounds great 👍

src/components/ReminderCtaButton.tsx
KaliedaRik: This is a controversial change I've introduced. The Epic includes the message because that's the message that shows up after user clicks on reminder button in RRCP banner/epics - but in BrazeLand the reminder button is always going to be a single-click action. I cleared adding the copy with Marketing in chat when doing the work to update the Epic reminder to the RRCP way-of-things. But when adding the reminder button to the Banner we run into the vertical space issue - the copy under the reminder button can take 2-4 lines of additional vertical space!

`hidePrivacyText` sounds like a good solution for the copy on both epic and banner! It does add an extra attribute which Marketing will need to be aware of when building epic/banner messages in Braze canvases - another thing that can be fixed through education?

I'll give it a go.

src/StyleableBannerWithLink/index.tsx
KaliedaRik: I'll experiment and reach out if (when) it starts taking too long to rectify

src/components/ReminderCtaButton.tsx
tjmw: Oh sorry that's actually not quite what I meant but is definitely an alternative & a valid approach! I just meant rename the current prop in this specific component to `shouldHidePrivacyText` or similar. Allowing marketing to configure as you suggested would definitely give the most flexibility. Perhaps worth asking around first in case there's a data privacy requirement to _always_ show this text? I'm just thinking that if we should always be showing it then it wouldn't be worth making it configurable.

tjmw: (But yes I also totally see your point about spacing on the banner!)

src/StyleableBannerWithLink/index.tsx
KaliedaRik: Fixed (I think)

src/components/ReminderCtaButton.tsx
KaliedaRik: Done

src/StyleableBannerWithLink/index.tsx
```diff
@@ -74,14 +82,36 @@ const StyleableBannerWithLink: React.FC<Props> = (props: Props) => {
...
+    const brazeProps = props.brazeMessageProps;
+
+    const styles = selfServeStyles(brazeProps, defaultColors);
+
+    const reminderCtaStyles = {
+        styleReminderButton: brazeProps?.styleReminderButton as OptionalColorValueHex,
+        styleReminderButtonBackground:
+            brazeProps?.styleReminderButtonBackground as OptionalColorValueHex,
+        styleReminderButtonHover: brazeProps?.styleReminderButtonHover as OptionalColorValueHex,
+        styleReminderAnimation: brazeProps?.styleReminderAnimation as OptionalColorValueHex,
```
tjmw, line 105: Ah sorry, this doesn't quite address my concern about the casting here. I think we're telling TypeScript something here which isn't necessarily true - that the value from Braze will either be empty (`undefined`) or a valid hex colour code. But I don't think that's correct - it could be anything - so we lose safety which could be provided by TypeScript. We do have a type guard `stringIsColorValueHex` which can safely take us from a `string` to a `ColorValueHex` as long as the value is valid. What happens if you remove the casting here and let the data flow through as a `string | undefined`? I think at some point downstream we validate that the value is a valid hex string, so perhaps that could be the point at which we go from a `string` to a `ColorValueHex` (and use the default if the value provided isn't valid). What do you think?

src/components/ReminderCtaButton.tsx
tjmw: So now we're making the privacy text optional and controllable from Braze for both banners and epics? If that's okay data-privacy wise then that sounds good 👍 I'd be interested in which the usual case is (to hide it or to show it), if we should usually show it I wonder if that should be the default? I.e. should the prop be `shouldHidePrivacyText` and that defaults to `false` (i.e. show the text) if omitted in the key value pairs.

KaliedaRik: In the context of a Braze banner/epic, where the reader has already signed in and thus (in theory) already knows we have their email address. I'm not convinced we need to show the privacy notice in all cases? I assume Braze itself can handle winnowing out users who ask for a reminder but have also set their communication preference to not receiving email. But (as always) I'm not a data privacy expert ...

I've set the default to `false` for epics. I've set it to `true` for banners, mainly because of the need to minimise the banner's vertical size.

src/StyleableBannerWithLink/index.tsx
KaliedaRik: Removing the casting leads to a TS error when I attempt to pass the CTA styles to the CTA component:

Screenshot 2023-11-22 at 09 34 35

Do you want me to spend more time investigating possible alternative solutions to this? Personally, I think in-browser testing all user input for these attributes using `stringIsColorValueHex()` is sufficient to prevent breaking the page CSS in production, and any additional TS won't give us much more beyond code documentation.

.changeset/ninety-mails-begin.md
```diff
@@ -0,0 +1,5 @@
+---
+'@guardian/braze-components': minor
```
tjmw, line 2: 👍

src/Epic/index.stories.tsx
```diff
@@ -55,6 +55,11 @@ export default {
             type: { name: 'string', required: false },
             description: 'Extra data to be associated with a reminder sign-up',
         },
+        showPrivacyText: {
+            name: 'showPrivacyText',
+            type: { name: 'string', required: false },
+            description: 'Set value to "true" to display the privacy text under the Reminder CTA',
```
tjmw, line 61: Is it worth mentioning the default behaviour here? I _think_ you said on epics the default is to show the privacy text?

tjmw (approved):
Great!

PR #444, opened 2023-12-12
Fix error when `createReminder` is called

tjmw:
## What does this change?

While attempting to test the reminders work from #433 and #422 in CODE I ran into an issue which was preventing the reminder sign-up from working. In the code we attempt to check whether we're running in Storybook, in which case we return a stub promise which immediately resolves. However, in the context of DCR `process.env` isn't available so the attempt to read `process.env.STORYBOOK` always fails and causes an error, resulting in the attempt to create the reminder failing. This change allows us to detect storybook in a way that's safe to do in DCR.

## How to test

I've tested locally running DCR and publishing braze-components with yalc. I see that the reminder signup from the banner now works:

Screenshot 2023-12-12 at 15 59 35

## Accessibility

### DIFF
--- a/.changeset/pink-apes-destroy.md
+++ b/.changeset/pink-apes-destroy.md
@@ -1,0 +1,5 @@
+---
+'@guardian/braze-components': minor
+---
+
+Fix createReminder function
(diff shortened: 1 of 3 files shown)


tomrf1 (approved):
👍

PR #446, opened 2023-12-27
Create a new self-serve newsletter epic

KaliedaRik:
## What does this change?
Currently all newsletter epics are hardcoded, which limits how Marketing colleagues can use such epics to promote newsletters to readers.

This PR introduces a new newsletter epic - **EpicNewsletter_SelfServe** - which gives them the ability to create an epic message, with a single-click sign-up CTA button, for any newsletter they need to promote.


src/BrazeEndOfArticleComponent.ts
```diff
@@ -58,8 +58,6 @@ const END_OF_ARTICLE_MAPPINGS: ComponentMapping<
     [EPIC_NAME]: Epic,
     [EPIC_WITH_HEADER_IMAGE_NAME]: EpicWithSpecialHeader,
 
-    [NEWSLETTER_EPIC_NAME]: NewsletterEpic,
```
tjmw, line 61: Interestingly, I think this component (which the specific newsletter epic components use under the hood) has the same functionality & interface as the new `EpicNewsletter_SelfServe`. The difference is that the generic `NewsletterEpic` functionality wasn't ever exposed in storybook, but I think that's the only reason it wasn't/couldn't be used by marketing. I think we could combine the work you've done here to surface things in storybook with the existing `NewsletterEpic` component, rather than having two similar (identical?) components: `NewsletterEpic` and `EpicNewsletter_SelfServe`. So the same generic component can either be used directly by marketing, for full flexibility, and continue to be used by the specific newsletter epic components (e.g. `AUNewsletterEpic`) for the time being. At some point I think we should deprecate those and the generic component should be used in all cases.

KaliedaRik: My initial attempt at this work was to create a `stories.tsx` file in the `NewsletterEpic` folder. It was by far the simplest solution ... but for the life of me I couldn't get the epic to display in Storybook.

This solution is uglier, but it works.

Thoughts?

tjmw: Ah right yeah that's what I was suggesting. Weird that you couldn't get it to display 🤔  I'm not sure why that would be but happy to pair if it'd be useful to try and figure it out?

src/NewsletterEpic/index.stories.tsx
```diff
@@ -0,0 +1,113 @@
...
+    },
+};
+
+const StoryTemplate = (
+    args: BrazeMessageProps & {
+        componentName: string;
+        newsletterId: string;
+    },
+): ReactElement | null => {
+    const imageUrl = grid(
```
tjmw, line 68: Does the grid plugin generate images which are appropriately sized for the Braze epic? (I _think_ this is the first use case for it in the epic?)

tjmw (approved):
Looks good!

PR #456, opened 2024-01-17
Bump peer dependencies to match DCR

tjmw:
## What does this change?

This bumps the peer dependency versions for this package (and dev dependencies to match).

See also guardian/dotcom-rendering#10203.

## Accessibility


tjmw:
This change has resulted in quite a few (fairly minor) Chromatic diffs. I'll investigate.

KaliedaRik (approved):
A nice bit of housekeeping!

tjmw:
> This change has resulted in quite a few (fairly minor) Chromatic diffs. I'll investigate.

After some investigation I found that the changes all seem to be related to line-height, which was slightly tightened up in a recent source-foundations release (which we're now using here). See https://github.com/guardian/csnx/blob/main/libs/%40guardian/source-foundations/CHANGELOG.md#1300.

On top of this the line-height now matches the RRCP epic and presumably other copy on the website.

Actually, since source-foundations is a peer dependency we're already rendering on the website using the new version and therefore already using the reduced line height so this PR doesn't make any material change to the way things are rendered on www.theguardian.com.

PR #459, opened 2024-01-30
Add a new styleable Newsletter banner component

KaliedaRik:
The purpose of this PR is to create a new `StyleableBannerNewsletter` banner component. Much as StyleableBannerWithLink extends self-serve design to BannerWithLink, so this new component extends BannerNewsletter. Users will be able to set the color requirements, and image, of a newsletter banner when they create the canvas for the newsletter campaign.

__Key gains:__ Braze newsletter banners often fail to match the color styling of wider newsletter promotional campaigns. This PR fixes that.

### Navigating the PR

Two new newsletter-related components have been created in the `src/components` folder, with functionality added to let them handle user-specified colors:
+ `NewsletterFrequencyBlock.tsx`
+ `NewsletterCtaButton.tsx`

The `newsletterCommon/index.ts` file, which previously hosted these components, has been deleted.

A new `StyleableBannerNewsletter` folder has been created , holding the following new files. The structure and functionality of the files are in line with other Banner/Epic components:
+ `index.tsx`
+ `canRender.ts`
+ `index.stories.tsx`

To better handle self-serve banner styling, the `styles/colorData.ts` file has been reorganised to more logically serve TS types/interfaces across the estate. 

Also, my previous decision to centralise banner CSS generation in `styles/bannerCommon.ts` was a mistake (too much difficulty understanding what the code was doing) - the file has been deleted, with the code relocalised into components.

Finally, the PR creates a new `BannerCloseButton.tsx` component in the `components` folder, and the (now 5) banner components updated to use it. The close button component localises all tracking and accessibility functionality into the component, and greatly simplifies the main banner components' code.


src/BannerNewsletter/index.tsx
```diff
@@ -110,39 +72,36 @@ const BannerNewsletter: React.FC<Props> = (props: Props) => {
...
+                        {secondParagraph && <p css={styles.secondParagraph}>{secondParagraph}</p>}
+                    </div>
+                    <NewsletterCtaButton
                         subscribeToNewsletter={subscribeToNewsletter}
                         newsletterId={newsletterId as string}
-                        ophanComponentId={ophanComponentId}
+                        ophanComponentId={ophanComponentId as string}
                         trackClick={trackClick}
+                        newsletterCta="Sign up"
+                        colors={defaultNewsletterCtaButtonColors}
```
tjmw, line 92: If these are the default colours, should we need to pass them in? What if they lived in the component itself and were used when no overrides were provided?

src/BannerWithLink/index.tsx
```diff
@@ -91,37 +63,25 @@ const BannerWithLink: React.FC<Props> = (props: Props) => {
...
-                            })
-                        }
-                    >
-                        {buttonText}
-                    </LinkButton>
+                    <PrimaryCtaButton
+                        buttonText={buttonText as string}
+                        buttonUrl={buttonUrl as string}
+                        showPaymentIcons={false}
+                        ophanComponentId={ophanComponentId as string}
```
tjmw, line 70: Again, in the interest of keeping interfaces simple, instead of passing through the `ophanComponentId` only to pass it to the `trackClick` callback what if we collapsed this and did the wiring outside? So drop the `ophanComponentId` arg and update `trackClick` to be `() => trackClick(ophanComponentId)` or similar? Actually it looks like `trackClick` also takes the `internalButtonId` which is defined in the `PrimaryCtaButton` component so it'd have to be provided, so the callback passed here would be something more like `(internalButtonId) => trackClick({ internalButtonId, ophanComponentId })`

tjmw (commented):
Hey this is looking good! I've not completed the review yet but submitting feedback so far to see what you think. I do wonder if some of the refactorings in here could be pulled out to a separate preliminary PR (or PRs) and then the work to add the new banner component could build on top of those? Just because there are quite a lot of changes in here and it's a little bit hard to see how everything fits together.

KaliedaRik:
I'll refactor into separate PRs

PR #462, opened 2024-02-09
Add new BannerCloseButton component

KaliedaRik:
## What does this change?
Moves the close button functionality and layout currently spread across the banner components into a single `BannerCloseButton` component.

## How to test
Storybook

## Images
No changes in display - check chromatic for differences


KaliedaRik:
This is the first PR for the work falling out of (the now closed) PR https://github.com/guardian/braze-components/pull/459 to create a new styleable Newsletter banner

src/AppBanner/index.tsx
```diff
@@ -138,28 +115,75 @@ export const AppBanner = (props: Props): ReactElement | null => {
...
-                                onClick={(e) => onCloseClick(e, CLOSE_BUTTON_ID)}
-                                tabIndex={0}
-                            >
-                                {' '}
-                            </Button>
-                        </ThemeProvider>
-                    </div>
+                    <BannerCloseButton
+                        trackClick={trackClick}
+                        setShowBanner={setShowBanner}
```
tjmw, line 124: Copied from #459:

> What do you think about simplifying the interface to BannerCloseButton and passing through a single callback to invoke when the button is clicked? I think that could be nicer as there's less coupling between the concerns of this component and more generic BannerCloseButton.

To clarify, what if we passed through a callback which was invoked by `BannerCloseButton` when clicked which did both click tracking and toggling the banner visibility? It could have the same signature as the `onCloseAction` method in `BannerCloseButton`. So the button ID is supplied by `BannerCloseButton` (something it knows about) but toggling visibility and `trackClick` are something the parent knows about, so the detail isn't leaked through to `BannerCloseButton`.

src/components/BannerCloseButton.tsx
```diff
@@ -0,0 +1,127 @@
...
+};
+
+export const BannerCloseButton: React.FC<BannerCloseButtonProps> = (
+    props: BannerCloseButtonProps,
+) => {
+    const {
+        trackClick,
+        setShowBanner,
+        ophanComponentId,
+        colors = defaultBannerCloseButtonColors,
```
tjmw, line 50: AFAICT the defaults are never used since the calling components always pass these (which I think makes sense - I wonder if it's possible to have default colours since this is used in different contexts). Perhaps the default should be removed and the prop becomes required?

src/AppBanner/index.tsx
```diff
@@ -98,27 +64,38 @@ export const AppBanner = (props: Props): ReactElement | null => {
...
+    };
+
+    const onCloseAction = (internalButtonId: number): void => {
+        setShowBanner(false);
+        document.body.focus();
+        trackClick({
+            internalButtonId,
+            ophanComponentId: ophanComponentId as string,
+        });
+    };
```
tjmw, line 79: I think these aren't used currently? (But I think they should be! See comment about passing an aggregated callback through to `BannerCloseButton` instead of more granular `trackClick` and `setShowBanner`).

tjmw (approved):
Looking good! Thank you for splitting this out. I left a few comments/questions but nothing major, see what you think.

src/components/BannerCloseButton.tsx
KaliedaRik: Good spot. All the static banners should be using the default colors and thus don't need to pass them back into the component. This has been fixed.

src/AppBanner/index.tsx
KaliedaRik: They're being used by both App Banner CTAs - they're separate from the the top-right "close" button but, unlike any other CTA button they actually do close the banner! These functions have been removed in their entirety from all the other banners and localised into the BannerCloseButton component ... does that make sense?

Screenshot 2024-02-09 at 17 19 36

src/AppBanner/index.tsx
KaliedaRik: Sorry - I'm not sure I follow. Closing the banner, and tracking anything in the banner that the user has interacted with, are different things. I'm not keen on combining them in a single callback - which I assume would have to be hardcoded into every banner component?

tjmw: Sorry I didn't explain well. You're right that they're separate things, but as far as the close button component is concerned maybe it's better to keep it "dumb" - I think all it should know about is "I've got a callback to execute when I get clicked". The callback is provided by the parent component which _does_ know about setting state and tracking etc but the child component can be blissfully ignorant, all it knows is I've got to execute something when I'm clicked.

KaliedaRik: Ah. I see where you're coming from. The way I've built the close button component ... it's not living in blissful ignorance: it knows that its purpose is to close a banner that includes it. It handles both closing the banner (via clicking on it, and by user pressing the ESC key), and reporting that user has interacted with it. It's the banner that doesn't care about tracking the close action, because it includes the close button component which handles all that for it.

The App banner component is the only banner which has to be different, in that it will also close when the user clicks on either of its CTA buttons. Hence the need to duplicate the onCloseClick and onCloseAction functions in this banner's code. 

Does that make sense?

src/AppBanner/index.tsx
tjmw: Ah I see, thanks for explaining 👍

PR #463, opened 2024-02-09
Move newsletter-related components into the components folder

KaliedaRik:
## What does this change?
Moves the newsletter frequency and CTA components into the components folder.

## How to test
Storybook

## Images
No changes in display - check chromatic for differences


KaliedaRik:
This is the second PR for the work falling out of (the now closed) PR https://github.com/guardian/braze-components/pull/459 to create a new styleable Newsletter banner

.changeset/serious-baboons-mate.md
```diff
@@ -0,0 +1,5 @@
+---
+'@guardian/braze-components': minor
```
tjmw, line 2: Maybe even `patch` - I don't think this will be visible to consumers at all? (Although I do like this changeset name 😂 )

src/NewsletterEpic/index.tsx
```diff
@@ -12,9 +12,71 @@ import {
...
+                        ophanComponentId={ophanComponentId}
+                        trackClick={trackClick}
+                        newsletterCta="Sign up"
+                    />
+                </div>
+            </section>
+        </ThemeProvider>
+    );
+};
+
```
tjmw, line 79: Git isn't doing a good job of showing the differences here - is this just a move of a large chunk of code or are there changes in here?

tjmw (approved):

src/NewsletterEpic/index.tsx
KaliedaRik: The only thing I did beyond importing and using the components (with their updated names) was move the styling to the bottom of the file ... Bad GitHub!

.changeset/serious-baboons-mate.md
KaliedaRik: I'm going to push both PRs through together so it should amount to just one `minor` bump I think

PR #465, opened 2024-02-20
Create new styleable newsletter banner

KaliedaRik:
## What does this change?
+ Add a new styleable banner newsletter component
+ Update existing banner newsletter component to use new component
+ Update existing banner with link component to use existing styleable banner with link component
+ Fix a bug where components were expecting to receive unnecessary functions
+ Update storybook stories

### Navigating the PR changes
The PR touches 12 files, as follows:
+ __AppBanner/index.tsx__ - updated to take into account changed color types, and localising the CSS for the primaryButton div
+ __BannerNewsletter/index.tsx__ - updated to make direct use of the new `StyleableBannerNewsletter` component
+ __BannerWithLink/index.tsx__ - updated to make direct use of the `StyleableBannerWithLink` component
+ __BrazeBannerComponent.ts__ - add the new `StyleableBannerNewsletter` component; change requirement to have `subscribeToNewsletter` and `fetchEmail` functions (to simplify Storybook stories)
+ __StyleableBannerNewsletter/*__ - new folder / 3 files required for the new component
+ __StyleableBannerWithLink/canRender.ts__ - removed check for imageAltText
+ __StyleableBannerWithLink/index.tsx__ - updated to take into account changed color types; and to handle CSS generation for this component (code previously in `styles/bannerCommon.ts`
+ __logic/canRender.ts__ - updated to include new component
+ __styles/bannerCommon.ts__ - deleted to simplify code base. Code to handle CSS generation for banners has been moved into `StyleableBannerWithLink/index.tsx` and `StyleableBannerNewsletter/index.tsx` files
+ __styles/colorData.ts__ - updated color interfaces for banners

The following Storybook stories (4 files) have been updated to remove unnecessary passing of `mockFetchEmail` and `mockSubscribe` functions to components that don't use them. Note that DCR supply both of these functions when calling braze-components; if a function is not passed through to a component that uses it, the component will refuse to render: 
+ __AppBanner/index.stories.tsx__
+ __BannerNewsletter/index.stories.tsx__
+ __BannerWithLink/index.stories.tsx__
+ __StyleableBannerWithLink/index.stories.tsx__

### UI tests
+ One new test for the new component
+ 5 tests required approval for changes - reminder month has moved from March to April

## How to test
+ Storybook
+ We will also test in DCR storybook as part of the related DCR PR to update the braze-components dependency


src/BrazeBannerComponent.ts
```diff
@@ -20,12 +20,16 @@ import {
...
+    StyleableBannerNewsletter,
+} from './StyleableBannerNewsletter';
 
 export type CommonBannerComponentProps = {
     componentName: string;
     brazeMessageProps: BrazeMessageProps;
-    subscribeToNewsletter: NewsletterSubscribeCallback;
-    fetchEmail: FetchEmail;
+    subscribeToNewsletter?: NewsletterSubscribeCallback;
+    fetchEmail?: FetchEmail;
```
tjmw, line 32: I don't think we should make these optional. In the current model there's a single "factory" component  - `BrazeBannerComponent` through which all other components are rendered. For example, DCR only knows about this component and not the different banner components underneath. So from this component's perspective I think these should _always_ be required, since from a types perspective we can't know for sure that they won't be needed. This may change when components are moved to DCR, but for now what do you think about putting this back to how it was? If these were made optional, they could be removed from the DCR call to `BrazeBannerComponent` and TS would be happy, but it would break the functionality around subscribing to newsletters or subscribing to reminders.

src/AppBanner/index.tsx
```diff
@@ -39,7 +39,8 @@ export const AppBanner = (props: Props): ReactElement | null => {
         trackClick,
     } = props;
 
-    const styles = selfServeStyles(props.brazeMessageProps, defaultColors);
+    const bannerWithLinkBaseColors = getColors(props.brazeMessageProps, defaultColors);
```
tjmw, line 42: For this banner I don't think `props.brazeMessageProps` ever contains colour props. I wonder if it would be clearer to pass an empty object `{}`? Or maybe even better update `getColors` to make the override colour props optional?

src/BannerNewsletter/index.tsx
```diff
@@ -1,123 +1,32 @@
...
-
-        ${from.tablet} {
-            ${headline.medium({ fontWeight: 'bold' })};
-            max-width: 100%;
-        }
-    `,
-    bold: css`
-        font-weight: bold;
-    `,
+    return <StyleableBannerNewsletter {...props}></StyleableBannerNewsletter>;
```
tjmw, line 31: 👏  Love this!

src/BannerWithLink/index.tsx
```diff
@@ -28,75 +26,12 @@ export const BannerWithLink: React.FC<Props> = (props: Props) => {
...
-    const styles = selfServeStyles(props.brazeMessageProps, defaultColors);
+    const { boldText } = props.brazeMessageProps;
 
-    const [showBanner, setShowBanner] = useState(true);
-
-    if (!showBanner) {
-        return null;
+    // StyleableBannerWithLink uses `highlight` instead of `boldText` attribute - this check fixes the discrepency so Marketing won't have to change existing canvases
+    if (boldText) {
+        props.brazeMessageProps.highlight = boldText;
```
tjmw, line 33: I wonder if rather than mutating `props.brazeMessageProps`, this could be resolved in the call to `StyleableBannerWithLink` below? That feels slightly cleaner to me (in a general feeling of mutation being best avoided where possible). I.e. instead of spreading the props with `...props` specify them explicitly.

src/BannerWithLink/index.tsx
```diff
@@ -1,18 +1,16 @@
...
 
+import { canRender, COMPONENT_NAME } from './canRender';
 export { COMPONENT_NAME };
 
 export type BrazeMessageProps = {
     ophanComponentId?: string;
     header?: string;
     body?: string;
     boldText?: string;
+    highlight?: string;
```
tjmw, line 13: It feels slightly odd that this has been added alongside `boldText`. Below it looks like we'll use `boldText` and pass it to `StyleableBannerWithLink` as highlight. So should it be part of the interface of this component?

src/StyleableBannerNewsletter/index.tsx
```diff
@@ -0,0 +1,299 @@
...
+};
+
+export type Props = {
+    brazeMessageProps: BrazeMessageProps;
+    trackClick: TrackClick;
+    subscribeToNewsletter?: NewsletterSubscribeCallback;
+};
+
+const StyleableBannerNewsletter: React.FC<Props> = (props: Props) => {
+    if (!props.subscribeToNewsletter || !canRender(props.brazeMessageProps)) {
```
tjmw, line 56: I think if we go back to making `subscribeToNewsletter` required then we won't need the left side of this condition.

src/StyleableBannerNewsletter/index.tsx
```diff
@@ -0,0 +1,299 @@
...
+
+    const styles = getBannerNewsletterStyles(brazeProps, newsletterBaseColors);
+
+    const [showBanner, setShowBanner] = useState(true);
+
+    if (!showBanner) {
+        return null;
+    }
+
+    console.log(styles);
```
tjmw, line 91: Should probably remove this console.log!

src/BrazeBannerComponent.ts
KaliedaRik: For me, this is one of those situations where TS is getting in the way of the real code. I'll try to explain.

DCR passes through two functions to its "factory" component, which we then **need to check in real code** to make sure they have indeed been passed through before rendering the affected component or banner.

- For `StyleableBannerWithLink` this is the `fetchEmail` function used to set up a reminder when the user clicks on the "Remind me in ..." button. If for any reason the function is missing then that button won't render - but the rest of the banner will, as banners with links can be about anything, not just a contributions request.

- For `StyleableBannerNewsletter` the only CTA is the "Subscribe to newsletter" button. Without the `subscribeToNewsletter` function there's no point showing the user the banner (bad UX) so when it's absent, the banner returns `null` rather than some HTML.

Newsletter banners don't need to know about `fetchEmail`, while link banners don't need to know about `subscribeToNewsletters` (the App banner does its own thing and doesn't care about either).

The current TS markup says that both of those functions are required by all banner components. That is a lie (and, in my view, bad documentation). It's up to the components to check in **real code** that the functions they require have indeed been passed to them by DCR - which, as code living outside of this repository must be treated as unreliable (and a security risk - anyone can change the relevant DCR code without our knowledge). I've added these checks to the components as part of this PR to mitigate the risk of the components breaking the main site when user tries to invoke a non-existent function.

Happy to chat about this IRL if you disagree.

src/AppBanner/index.tsx
KaliedaRik: Updated

src/BannerWithLink/index.tsx
KaliedaRik: Updated

src/BannerWithLink/index.tsx
KaliedaRik: Agreed. Will remove it. Hopefully we can persuade Marketing to stop using this banner in the near future, then remove it from the repo.

src/StyleableBannerNewsletter/index.tsx
KaliedaRik: Please see long comment above

src/StyleableBannerNewsletter/index.tsx
KaliedaRik: Removed

src/BrazeBannerComponent.ts
tjmw: Hey thanks for writing this up! Yep totally understood that there may be bits that are sub-optimal at the moment but I don't think changing the interface to the _outer_ factory component is the way to address those things. You're right that the individual inner components don't always need all the props provided to the outer component but the consumer of the factory component doesn't/shouldn't have to be aware of that in the current setup I don't think. All the caller knows is, "I've got some props to pass from Braze which _might_ result in any of the underlying components being rendered". So it has to pass all the dependencies. I'm all for making sure the inner components' interfaces reflect reality. Regardless of how this conversation ends up perhaps this could be done in a separate PR (I don't think it's directly related to or depended on by the new banner component).

KaliedaRik: Reverted.

src/BannerWithLink/index.tsx
```diff
@@ -21,82 +19,22 @@ export type BrazeMessageProps = {
...
         brazeMessageProps: {
-            ophanComponentId,
-            header,
-            body,
-            boldText,
-            buttonText,
-            buttonUrl,
-            imageUrl,
+            ...props.brazeMessageProps,
+            highlight: props.brazeMessageProps.boldText,
```
tjmw, line 35: 👏

src/StyleableBannerWithLink/canRender.ts
```diff
@@ -7,13 +7,10 @@ import type { BrazeMessageProps } from './index';
...
-        brazeMessageProps;
+    const { header, body, buttonText, buttonUrl, imageUrl, ophanComponentId } = brazeMessageProps;
     if (imageUrl && !isImageUrlAllowed(imageUrl)) {
         console.log(`Image URL ${imageUrl} is not allowed`);
         return false;
     }
-    return Boolean(
-        header && body && buttonText && buttonUrl && imageUrl && imageAltText && ophanComponentId,
-    );
+    return Boolean(header && body && buttonText && buttonUrl && imageUrl && ophanComponentId);
```
tjmw, line 15: When imageAltText isn't provided, do we just not include any? Or is there a default? (I'm guessing we simply omit it).

src/StyleableBannerWithLink/index.tsx
```diff
@@ -112,7 +116,7 @@ export const StyleableBannerWithLink: React.FC<Props> = (props: Props) => {
                             colors={primaryCtaColors}
                             trackClick={trackClick}
                         />
-                        {reminderStage && (
+                        {fetchEmail && reminderStage && (
```
tjmw, line 119: The `fetchEmail` condition can probably go, since it's back to being required?

tjmw (approved):

src/StyleableBannerNewsletter/index.tsx
KaliedaRik: Apologies, but I'd rather delete the PR than remove the check.

PR #480, opened 2024-06-05
Fix Chromatic action

tjmw:
## What does this change?

I've started seeing errors for the Chromatic action:

`error @guardian/braze-components@20.0.0: The engine "node" is incompatible with this module. Expected version "^18.15 || ^20.9". Got "20.8.1"`

It seems that the Chromatic action does its own thing with regards to the Node version it uses. The version provided by the setup-node action isn't honoured by Chromatic. Even so, I've moved to using the standard setup-node action and away from the deprecated Guardian setup Node action.

I've slightly loosened the Node version string under `engines` in package.json, to reflect the version Chromatic seems to be using. This is the real fix to get things working.

Also bump the Chromatic action, which was way behind.

## Accessibility


rBangay (approved):
👍

PR #481, opened 2024-06-06
Run the main CI workflow for dependabot PRs

tjmw:
## What does this change?

Currently we skip the main CI workflow for dependabot builds, but we don't want to do that. It means we don't get tests run, Riff Raff artefacts generated etc. I think in the past this was required because of wanting to limit which secrets dependabot could access (there wasn't any fine-grained control) but now this can be configured separately.

I learned on chat that the growth team have already removed this condition for their repos, so let's do the same.

## Accessibility


rBangay (approved):
👍

PR #500, opened 2024-06-14
Upgrade rollup-plugin-filesize

tjmw:
## What does this change?

Upgrade `rollup-plugin-filesize` to 10.x.x. Even though this is a major version bump, it seems to work in a backwards compatible way - the file size reporting when running `yarn build` looks the same.

Doing this because it removes one of the paths causing us to pull in the ip package, which currently has an unpatchable vulnerability.

## Accessibility


yarn.lock
```diff
@@ -8805,10 +8769,13 @@ io-ts@^2.2.10:
   resolved "https://registry.yarnpkg.com/io-ts/-/io-ts-2.2.16.tgz#597dffa03db1913fc318c9c6df6931cb4ed808b2"
   integrity sha512-y5TTSa6VP6le0hhmIyN0dqEXkrZeJLeC5KApJq6VLci3UEKF80lZ+KuoUs02RhBxNWlrqSNxzfI7otLX1Euv8Q==
 
-ip@^1.1.5:
```
tjmw, line 8808: This is the bit we want gone!

PR #508, opened 2024-09-23
Use source 8

tomrf1:
DCR will [soon be upgrading](https://github.com/guardian/dotcom-rendering/pull/12342) to source 8.

The changes here are:
1. use `cssOverrides` instead of `css` prop
2. replace legacy foundations imports. I used these mappings to help with this: https://gist.github.com/jamesmockett/284f442762360e103c14bd83f90ba4dd


jamesmockett (approved):
Thanks for doing this Tom 🙏

PR #511, opened 2024-11-08
Add security policy property to CloudFront

MaelGNM:
## What does this change?

This adds the `securityPolicy` property to CloudFront for the hosted storybook, and sets it to the highest options `TLS_V1_2_2018`. This will stop it from allowing insecure TLS ciphers.

## How to test

Tested in CODE and PROD (as Grid integration doesn't work in CODE)


johnduffell (approved):
good 👍

PR #513, opened 2024-12-06
CDK cross-spawn vulnerability fix and dependency updates

rBangay:
## What does this change?
 
Major version update to "@guardian/cdk" alongside a new resolutions block in the package.json file on order to fix a vulnerability in cross-spawn.

Additional updates to 
- aws-cdk
- aws-cdk-lib
- constructs
- jest
- prettier
- ts-jest
- ts-node


MaelGNM (approved):
Looks good!

PR #515, opened 2025-03-19
patch vuln in path-to-regex (via express and storybook)

johnduffell:
This is intended to fix https://github.com/advisories/GHSA-rhx6-c78j-4q9w
which came via storybook->express->path-to-regex
It looks like the express dependency has been removed in the later version of storybook.
It also looks like we need a new storybook/test package, otherwise chromatic build fails with
```
=> Failed to build the preview
             SB_BUILDER-WEBPACK5_0002 (WebpackInvocationError): Module not found: Error: Can't resolve '@storybook/test' in './node_modules/@storybook/react/dist'
```

Looks ok on story book on CODE, and since the only changes are to devDependencies, then there's no need to release a new package to npm.


package.json
```diff
@@ -51,14 +51,15 @@
...
-    "@storybook/addon-viewport": "^8.3.6",
-    "@storybook/addon-webpack5-compiler-babel": "^3.0.3",
-    "@storybook/components": "^8.3.6",
-    "@storybook/react": "^8.3.6",
-    "@storybook/react-webpack5": "^8.3.6",
+    "@storybook/addon-controls": "^8.3.7",
+    "@storybook/addon-docs": "^8.3.7",
+    "@storybook/addon-toolbars": "^8.3.7",
+    "@storybook/addon-viewport": "^8.3.7",
+    "@storybook/addon-webpack5-compiler-babel": "^3.0.5",
```
johnduffell, line 58: I just bumped this to latest because I'd done all the other storybook packages

MaelGNM (approved):

PR #518, opened 2025-03-28
@braze/web sdk v3 to v4 to v5

andresilva-guardian:
## What does this change?

This pull request upgrades @braze/web-sdk-core from version 3.5.1 to @braze/web-sdk version 5.8.1. The change is reflected in the package and documented in the CHANGELOG.md. Additionally, updates have been made to jest.config.js to refine module transformation patterns. More detail could be read [here](https://trello.com/c/213tggX9/749-braze-v5-for-braze-components).

### Breaking changes ⚠️ 

The "cardId" and "campaignId" properties of In-App Messages were [removed](https://github.com/braze-inc/braze-web-sdk/blob/master/UPGRADE_GUIDE.md#removed-apis-1) on the newer versions of this SDK.
To avoid data loss, I'm moving them into "extras" property to keep the same data as before.

## How to test

1. Run npm install to ensure dependencies are updated.
2. Execute npm test to verify that all unit tests pass successfully.
3. Perform a manual test to ensure the Braze SDK integration continues to function as expected.

- No breaking changes in application functionality.
- All automated tests pass.
- Expected behavior of Braze-related features is maintained.

- The SDK upgrade may introduce breaking changes or deprecations. Testing should confirm compatibility.
- The Jest configuration changes should be verified to ensure they do not disrupt existing test coverage.

## Images

N/A

## Accessibility

- No direct changes affecting accessibility.
- Standard testing practices should be followed to confirm no unintended regressions.

### DIFF
--- a/src/logic/LocalMessageCache.tsx
+++ b/src/logic/LocalMessageCache.tsx
@@ -74,24 +74,13 @@ const hydrateMessage = (
      * https://github.com/braze-inc/braze-web-sdk/blob/master/UPGRADE_GUIDE.md#removed-apis-1
      * We will move them into "extras" to keep the same data as before
      */
-    let extras: Record<string, string> | undefined = messageData.extras;
-    if (messageData.cardId) {
-        if (!extras) {
-            extras = {};
-        }
-        extras.cardId = messageData.cardId;
-    }
-    if (messageData.campaignId) {
-        if (!extras) {
-            extras = {};
-        }
-        extras.campaignId = messageData.campaignId;
-    }
-
-    /* eslint-disable @typescript-eslint/no-unsafe-member-access */
     const hydratedMessage = new brazeInstance.HtmlMessage(
         messageData.message,
-        extras,
+        {
+            ...(messageData.extras || {}),
+            ...(messageData.cardId && { cardId: messageData.cardId }),
+            ...(messageData.campaignId && { campaignId: messageData.campaignId }),
+        },
         messageData.triggerId,
         messageData.dismissType,
         messageData.duration,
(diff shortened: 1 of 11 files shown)


src/logic/LocalMessageCache.tsx
```diff
@@ -65,14 +65,33 @@ const readQueue = (slotName: MessageSlotName): CachedMessage[] => {
...
+            extras = {};
+        }
+        extras.cardId = messageData.cardId;
+    }
+    if (messageData.campaignId) {
+        if (!extras) {
+            extras = {};
+        }
+        extras.campaignId = messageData.campaignId;
+    }
```
johnduffell, line 89: I'm not really a TS programmer but for simplicity is it worth doing something along these lines?
```suggestion
    const extras =  { 
      ...messageData.extras || {},
      ...messageData.cardId && { cardId: messageData.cardId },
      ...messageData.campaignId && { campaignId: messageData.campaignId }
    }
```

andresilva-guardian: Hey @johnduffell , thanks for your suggestion! That was my first approach. But the thing is, by doing `...messageData.extras || {},` we are taking a decision of initializing "extras" even if we don't need to. I was afraid of doing that thinking about the consequences, but since you also believe that it's ok, I will change it, thanks!

johnduffell: see what you mean, I didn't test that scenario.  I would have thought it would be ok.

Trying to handle it in TS turns into something that although shorter than the original it doesn't seem clear any more
```
    const extrasArray = [ 
      messageData.extras,
      messageData.cardId && { cardId: messageData.cardId },
      messageData.campaignId && { campaignId: messageData.campaignId }
    ].filter((v) => v !== undefined);

    const extras = extrasArray.length > 0 ? Object.assign({}, ...extrasArray) : undefined;
```

also see https://playcode.io/2322453

PR #522, opened 2025-05-14
Block public access to storybook S3 bucket

MaelGNM:
## What does this change?

This updates the cloudformation file for the Storybook S3 bucket so that the bucket blocks all public access. There is a global setting on the AWS account already blocking public access, but the fact the bucket itself isn't specifically blocking public access is reported as a vulnerability.
There is CloudFront in front of the bucket so the bucket doesn't need to allow public access.

## How to test

- [x] Deploy braze-components-storybook
- [x] Check that bucket now blocks public access
- [x] Check that storybook can be accessed in CODE
- [x] Check that storybook can be accessed in PROD

Once this goes out it should remediate one of the reported FSBP vulnerabilities.

## Images

| Before | After |
| --- | --- |
|  image | image |

## Note about deployment

The cloudformation is not part of CI/CD and is deployed manually. We may want to change that in future.

### DIFF
--- a/cloudformation-static-s3.yaml
+++ b/cloudformation-static-s3.yaml
@@ -33,6 +33,11 @@ Resources:
         Type: AWS::S3::Bucket
         Properties:
             BucketName: braze-components-storybook
+            PublicAccessBlockConfiguration:
+                BlockPublicAcls: true
+                IgnorePublicAcls: true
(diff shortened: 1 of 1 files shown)


rBangay (approved):
👍

PR #523, opened 2025-09-09
Refactor to Fix SSM Parameter Store for S3 bucket name in deployment and CloudFormation

ecguardian:
## What does this change?
When deploying braze-components PRs to CODE/PROD, Riff-Raff generates a warning message advising not to include S3 bucket names in the riff-raff.yaml file. 

This PR refactors the deployment and infrastructure setup to remove the hardcoded S3 bucket name from `riff-raff.yaml` and CDK code.

- Adds a CloudFormation parameter to reference the SSM Parameter Store entry for the bucket name.

- Updates riff-raff.yaml to use the secure parameter.

- Updates CDK code to use the parameter instead of a hardcoded value.

- Improves security and removes the Riff-Raff warning about exposed bucket names.

## Accessibility


johnduffell:
> "Updates riff-raff.yaml to use the secure SSM Parameter Store entry."

it doesn't have to be marked as a SecureString (if that's what you mean), as it is classed as "private" rather than "secret".  The benefit is it's a tiny bit easier to manage in the console, and also it costs a millionth of a penny less every time we use it.

cdk/lib/braze-components.ts
```diff
@@ -20,11 +21,16 @@ export class BrazeComponents extends GuStack {
     constructor(scope: App, id: string, props: BrazeComponentsStackProps) {
         super(scope, id, props);
 
-        const bucketNameFromStaticCloudformationStack = 'braze-components-storybook';
+        // Get bucket name from SSM Parameter Store
+        const bucketNameParameter = StringParameter.fromStringParameterName(
+            this,
+            'StorybookAssetsBucketParameter',
+            '/braze-components/storybook-assets-bucket',
```
johnduffell, line 28: the format of bucket keys (in membership account - but targeting is similar) is like this
image

I'm not sure why we go for "." instead of "-" - I think that's a riffraff thing, but should we be consistent?
```suggestion
            '/braze-components/storybook-assets.bucket',
```

cdk/lib/braze-components.ts
```diff
@@ -20,11 +21,16 @@ export class BrazeComponents extends GuStack {
     constructor(scope: App, id: string, props: BrazeComponentsStackProps) {
         super(scope, id, props);
 
-        const bucketNameFromStaticCloudformationStack = 'braze-components-storybook';
+        // Get bucket name from SSM Parameter Store
```
johnduffell, line 24: I think it should be self explanatory and comments often end up outdated or separated from their associated code
```suggestion
```

PR #545, opened 2026-05-13
Use pnpm

tomrf1:
- relaces yarn with pnpm
- removes version ranges from package.json


.github/workflows/braze-components.yml
```diff
@@ -53,20 +58,29 @@ jobs:
...
-        run: yarn install
+        run: pnpm install
 
       - name: Storybook build
-        run: yarn build-storybook
+        run: pnpm build-storybook
 
       - name: CDK
         working-directory: cdk
-        run: ./script/ci
```
tomrf1, line 69: I've deleted this script as we normally just inline it here

juabara (approved):

